PR #18, opened 2024-01-05
[Kunlun] Remove torch_xmlir from src

ZLkanyo009:


megatron/megatron/arguments.py
```diff
@@ -439,16 +435,6 @@ def validate_args(args, defaults={}):
     if args.sequence_parallel:
         args.async_tensor_model_parallel_allreduce = False
 
-    if torch_xmlir is None:
```
aoyulong, line 442: I think this should not be removed.

megatron/megatron/arguments.py
```diff
@@ -548,10 +534,7 @@ def core_transformer_config_from_args(args):
             kw_args[f.name] = getattr(args, f.name)
     kw_args['persist_layer_norm'] = not args.no_persist_layer_norm
     kw_args['layernorm_zero_centered_gamma'] = args.apply_layernorm_1p
-    if torch_xmlir is None:
-        kw_args['deallocate_pipeline_outputs'] = True
-    else:
-        kw_args['deallocate_pipeline_outputs'] = False
+    kw_args['deallocate_pipeline_outputs'] = False
```
aoyulong: I think `kw_args['deallocate_pipeline_outputs']` should be `True`.

megatron/megatron/text_generation/generation_single_thread.py
```diff
@@ -233,8 +226,7 @@ def generate_tokens_probs_and_return_on_first_stage_stream_sub_process(
                 # length, it means we have started generating tokens
                 started = lengths <= context_length
                 # Update the tokens.
-                if started:
-                    tokens[0, context_length] = new_sample[0]
+                tokens[started, context_length] = new_sample[started]
```
aoyulong: Why does this pr change these two lines? They are not related to the kunlun chip.

megatron/tools/run_text_generation_uvicorn_server_single_thread.py
```diff
@@ -179,7 +173,7 @@ def postprocessing(fun):
                 lock_stream.release()
 
             return StreamingResponse(trans(), media_type="text/plain",
-                                background=BackgroundTask(postprocessing, fun), lock=lock_stream)
+                                background=BackgroundTask(postprocessing, fun))
```
aoyulong, line 176: Same question as above.

ZLkanyo009: This issue was left over from before, and originally there was no lock=lock_stream added. Whether there is lock=lock_stream or not, the effect is the same.

PR #26, opened 2024-01-16
【mthreads】enable use-flash-attn switch

shang-mt:


megatron/tools/__init__.py
aoyulong, line 1: Why does this pr add a `__init__.py`?

PR #35, opened 2024-01-25
Add qkv bias flags and set wandb offline mode.

ftgreat:


megatron/megatron/arguments.py
```diff
@@ -1039,6 +1039,9 @@ def _add_training_args(parser):
     group.add_argument('--disable-bias-linear', action='store_false',
                        help='Disable bias in the linear layers',
                        dest='add_bias_linear')
+    group.add_argument('--enable-bias-linear-qkv', action='store_true',
```
aoyulong: We should follow the convention as `--disable-bias-linear`.

megatron/megatron/model/transformer.py
```diff
@@ -618,15 +618,15 @@ def __init__(self, config, layer_number,
                 query_projection_size,
                 config=config,
                 init_method=config.init_method,
-                bias=config.add_bias_linear,
+                bias=(args.add_bias_linear_qkv or args.add_bias_linear),
```
aoyulong: We should also support the new `config.add_bias_linear` instead of using the `args.add_bias_linear` directly since we will sync with the upstream.

PR #41, opened 2024-01-30
[DCU]Support flagscale and aquila2

yingzhao27:


megatron/megatron/arguments.py
```diff
@@ -930,7 +930,8 @@ def _add_training_args(parser):
                        'individual Transformer layers per pipeline stage and do the '
                        'rest without any recomputing at specified granularity'
                        'default) do not apply activations recompute to any layers')
-    group.add_argument('--recompute-num-layers', type=int, default=1,
+    # From dcu
+    group.add_argument('--recompute-num-layers', type=int, default=None,
```
aoyulong: I think we don't need to change this default value because this can be set if we want to use it.

megatron/megatron/fused_kernels/__init__.py
aoyulong, line 1: These changes may be not OK because the code also be shared by other vendors.

PR #64, opened 2024-03-13
Add mixtral pretrain

zhaoyinglia:


megatron/megatron/core/distributed/distributed_data_parallel.py
```diff
@@ -133,10 +133,10 @@ def allocate_grad_buffers_for_parameters(
                 grad_buffer_param_index_map[dtype] = grad_buffers[-1].param_index_map
 
             if grad_buffer_param_index_map:
-                for name, param in self.module.named_parameters():
-                    if param.requires_grad:
-                        dtype = torch.float if accumulate_allreduce_grads_in_fp32 else param.dtype
-                        self.param_name_to_index_map[name] = (tuple(param.shape), grad_buffer_param_index_map[dtype][param])
+                for param in params:
```
aoyulong, line 134: This seems to be wrong. The correct should be
```
if grad_buffer_param_index_map:
    for dtype, params in grad_dtype_to_params.items():
        for param in params:
```

zhaoyinglia: done.

aoyulong (approved):
LGTM

PR #108, opened 2024-04-30
add conf for llama2, add checkpoint convert for llama

phoenixdong:
- Add the conf for running llama2, and refer to the readme for running llama2 in Megatron for the data source.
- Add functionality to convert llama checkpoint back and forth between mcore and transformers.


flagscale/train/train_llama2.py
aoyulong, line 1: It's better to use the same entry point for different llama serials and keep the configuration file to be independent. So change the file name to `train_llama.py`.

aoyulong (commented):
Please update to llama3.

PR #133, opened 2024-06-06
Add gpu check

phoenixdong:
- The purpose is to avoid CI conflicts in multiple runners.
- The method is to check the GPU memory usage and wait until the memory is sufficient to start the test.


scripts/gpu_check.sh
aoyulong, line 1: Please relocate this to the tool folder or the tests folder.

PR #134, opened 2024-06-07
[Iluvatar] Support tensor parallel heterogeneous training

njuerect:
This PR adds support for training LLM on heterogeneous devices with different tensor parallel sizes.


flagscale/train/hetero/train_llama.py
```diff
@@ -0,0 +1,292 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.  All rights reserved.
```
zhaoyinglia, line 1: Do not copy a `train_llama.py`, which should be unified training entrance. You can put the functional changes that must be rewritten in hetero folder.

flagscale/train/hetero/training.py
```diff
@@ -0,0 +1,680 @@
+# Copyright (c) 2022, NVIDIA CORPORATION. All rights reserved.
```
zhaoyinglia, line 1: Same as the `train_llama.py` above.

flagscale/train/hetero/train_llama.py
aoyulong: Great idea! This will definitely be improved in the next pull request.

aoyulong (approved):
LGTM

PR #140, opened 2024-06-09
[AutoTuner]Optimize prune and fix some bugs

Caozhou1995:
In this PR, we optimize prune rules:

1. adding prune rules based on tp and pp
2. adding dist optimizer rules
3. adding Multi-Dimensional joint rules.
Pruning rules now cover dist optimizer, mbs, sp, recompute, tp, pp, and any combination between them


aoyulong (approved):
LGTM

PR #174, opened 2024-08-07
[NewFeature] Support Patch mechanism 

shenzhu1993:
This is a brand-new patch mechanism aimed at helping manufacturers and users better utilize the FlagScale framework for training and inference.

1. Homogeneous scenarios 
   Under this mechanism, manufacturers use the tools provided by FlagScale to automatically adapt the code between the base-commit-id and the current-commit-id, generating a compliant patch, automatically commit the code.  The sample command is as follows:
  
            cd FlagScale
            python tools/patch/patch.py --device-type A_X100 --base-commit-id aaaa --current-commit-id bbbb

            --device-type:  Required. Manufacturers name and hip mode.
            --base-commit-id: Required. Commit id modified based on FlagScale.
            --current-commit-id: Not required. The commit-id that want to  git commit, if it's None, the head commit id will be choosen.
     If manufacturers or user want to unpatch the code adapted, the sample command is as follows:
  
            cd FlagScale
            python tools/patch/unpatch.py --device-type A_X100 --commit-id aaaa --dir build

            --device-type:  Required. Manufacturers name and hip mode.
            --commit-id: Required. The base-commit-id Under the FlagScale's hardwares directory, in the manufacturer's directory. This is the --base-commit-id you enter when using patch.py.
            --dir: Not required. Directory used to store the adapted code. if it's None, The unpatch operation will be performed in the current directory.
     If you want to push the patch code,  each commit overwrites the previous patch, so there will be only one patch file at a time. Therefore, we recommend using the following command for pushing:
             
            git push  --force  origin HEAD:main
     
    Notice: If you have iterated development based on the same base_commit_id, you may need  to resolve conflicts before pushing, because the same patch file has been modified. We recommend using rebase to resolve them.
 
2. Heterogeneous scenarios

      Heterogeneous scenarios are as simple as homogeneous scenarios. You only need to add the manufacturer's name to the --device-type option. The sample command is as follows:

           cd FlagScale
           python tools/patch/patch.py --device-type A_X100 B_Y100 --base-commit-id aaaa --current-commit-id bbbb
              
           --device-type:  Required. Both manufacturers name and hip mode adapt to perform heterogeneous training or inference.
           --base-commit-id: Required. The commit-id that  git pull from FlagScale.
           --current-commit-id: Not required. The commit-id that want to  git commit, if it's None, the head commit id will be choosen.
      The input parameters have the same meaning as in homogeneous scenarios. However, it is important to note that the manufacturer's name and base-commit-id must exist in patch/hetero.txt. The format in patch/hetero.txt is as follows:
         
           aaaa: A_X100  B_Y100
      If manufacturers or user want to unpatch the code adapted. The sample command is as follows:
          
          cd FlagScale
          python tools/patch/unpatch.py --device-type A_X100 B_Y100 --commit-id aaaa --dir build
           
            --device-type:  Required.Manufacturers name and hip mode 
            --commit-id: Required. The base-commit-id Under the FlagScale's hardwares directory, in the manufacturer's directory
            --dir: Required. Directory used to store the unpatch code.


patch/common.py
```diff
@@ -0,0 +1,95 @@
...
+    check_path()
+    try:
+        repo = Repo(path)
+    except:
+        raise FileNotFoundError(path)
+    assert not repo.bare
+    return repo
+
+
+def crete_tmp_dir(dir_path=None, tmp_str=None):
```
zhaoyinglia: is tmp dir needed to be saved for display? Is not, you can use [tempfile](https://docs.python.org/3/library/tempfile.html) to manage.

patch/common.py
```diff
@@ -0,0 +1,95 @@
+import re
```
zhaoyinglia: rename this filename as `utils.py`

patch/patch.py
```diff
@@ -0,0 +1,317 @@
...
+    git_init,
+    save_patch_to_tmp,
+    check_branch_name,
+)
+
+path = os.getcwd()
+
+
+def _add_auto_generate_args():
+    """Set input argument."""
```
zhaoyinglia: put the argparse in `utils.py` or create a new file `argument.py`?

zhaoyinglia (commented):
Put this patch dir into tools. Add a README.md in patch dir, and copy this pr's description to the readme file.

patch/patch.py
shenzhu1993: Because the argument is not too much , we didn't create a new utils.py or argument.py.

patch/common.py
shenzhu1993: OK ，I will try the tempfile in next commit.

patch/common.py
shenzhu1993: Thanks.

requirements/requirements.txt
```diff
@@ -13,6 +13,7 @@ ray >= 2.9
 matplotlib
 py3nvml
 graphviz
+gitpython
```
heavyrain-lzy, line 16: Is it better to place this tool into file `requirements-dev.txt`?

PR #175, opened 2024-08-08
[CI] Update test scripts, unit tests, and test environments.

phoenixdong:
- Update test scripts

1. If an exception occurs in the test, exit the test immediately and throw the exception.
2. Print test progress.

- Update unit tests

1. Solve the problem of distributed initialization.
3. Solve the synchronization problem when creating temporary files.
4. Solve the problem of computing power mismatch.
5. Delete redundant methods.

- Update test environments

1. Install new dependency packages.
2. Container mounted shared memory.


aoyulong (approved):
LGTM

PR #180, opened 2024-08-12
[Hetero] Improve the general heterogeneous training parallelism

Corle-hyz:
This PR enables heterogeneous training based on process meshes. Different tensor parallel sizes, pipeline parallel splits, and multi-meshes are allowed.

Usage
`python run.py --config-path ./examples/aquila/conf --config-name config_hetero`

### DIFF
--- a/flagscale/train/hetero/parallel_context.py
+++ b/flagscale/train/hetero/parallel_context.py
@@ -315,10 +315,9 @@ class ProcessMesh:
         else:
             parallel_sizes = self._rank_generator.ordered_size_wo_ep
             order = self._rank_generator.order_wo_ep
+        order = order.split("-")
         if token in order:
-            # index() method return the start pos of token in tp-cp-dp-pp
-            # eg. if token==dp, then the index() method will return 6, while we need 2 actually
-            return parallel_sizes[order.index(token)//3]
+            return parallel_sizes[order.index(token)]
         else:
             raise ValueError(f"Invalid token: {token}")
 


flagscale/train/hetero/p2p_communication.py
```diff
@@ -17,10 +17,76 @@
...
+from flagscale.train.hetero.parallel_context import ParallelContext
 # Types
 Shape = Union[List[int], torch.Size]
 
+def warm_up_comm_group_hetero(config: ModelParallelConfig):
+    group = None
+    rank = torch.distributed.get_rank()
+    para_ctx = get_parallel_context()
+    pp_groups = para_ctx.get_pipeline_model_parallel_group()
+    tensor_shape = [1, 3]
```
aoyulong: Why was the shape [1, 3] chosen?

flagscale/train/hetero/parallel_context.py
```diff
@@ -317,7 +316,9 @@ def get_parallel_size(self, token, independent_ep=False):
             parallel_sizes = self._rank_generator.ordered_size_wo_ep
             order = self._rank_generator.order_wo_ep
         if token in order:
-            return parallel_sizes[order.index(token)]
+            # index() method return the start pos of token in tp-cp-dp-pp
+            # eg. if token==dp, then the index() method will return 6, while we need 2 actually
+            return parallel_sizes[order.index(token)//3]
```
aoyulong: "What is the reason for index() returning 6 when the token is 'dp'?"

flagscale/train/hetero/parallel_context.py
```diff
@@ -450,15 +455,16 @@ def is_initialized(self):
...
-        for tp, cp, ep, dp, pp in self._args.hetero_process_meshes:
+        # No ep for now
+        # for tp, cp, ep, dp, pp in self._args.hetero_process_meshes:
+        for tp, cp, dp, pp in self._args.hetero_process_meshes:
             process_mesh = ProcessMesh(
                 tensor_model_parallel_size=tp,
                 context_parallel_size=cp,
                 data_parallel_size=dp,
                 pipeline_model_parallel_size=pp,
-                expert_model_parallel_size=ep,
```
aoyulong: Why was this argument removed?

flagscale/train/hetero/parallel_context.py
```diff
@@ -508,28 +514,31 @@ def build_inter_mesh_process_groups(self, process_mesh1, process_mesh2):
...
                     sp_dim, dp_dim, pp_dim = dst_coord
-                    dst_coord = (sp_dim % tp2, sp_dim // tp2, dp_dim, pp_dim)
+                    dst_coord = [sp_dim % tp2, sp_dim // tp2, dp_dim, pp_dim]
                     dst_rank = process_mesh2.logical_coords_to_physical_ranks(
                         [dst_coord]
                     )[0]
                     ranks = [src_rank, dst_rank]
                     timeout = max(process_mesh1._timeout, process_mesh2._timeout)
                     group = torch.distributed.new_group(ranks, timeout=timeout)
-                    if self._rank in [src_rank, dst_rank]:
```
aoyulong, line 531: Why was the if statement removed?

megatron/megatron/core/pipeline_parallel/p2p_communication.py
```diff
@@ -127,35 +129,36 @@ def _batched_p2p_ops(
     group: torch.distributed.ProcessGroup
 ):
     ops = []
+
     if tensor_send_prev is not None:
         send_prev_op = torch.distributed.P2POp(
             torch.distributed.isend,
             tensor_send_prev,
-            get_pipeline_model_parallel_prev_rank(),
+            get_pipeline_model_parallel_prev_rank(group=group),
```
aoyulong, line 137: Ensure that the same functions of _p2p_ops are explicitly assigned to the group.

flagscale/train/hetero/p2p_communication.py
```diff
@@ -17,10 +17,76 @@
...
+    if comm_with_front_layer:
+        total_prev_pipeline_model_parallel_size = 0
+        for i in range(0, para_ctx._current_process_mesh_index):
+            total_prev_pipeline_model_parallel_size += para_ctx._process_meshes[i]._rank_generator.pp
+        return core.parallel_state.get_pipeline_model_parallel_rank() == total_prev_pipeline_model_parallel_size
+    else:
+        total_current_pipeline_model_parallel_size = 0
+        for i in range(0, min(para_ctx._current_process_mesh_index + 1, len(para_ctx._process_meshes))):
+            total_current_pipeline_model_parallel_size += para_ctx._process_meshes[i]._rank_generator.pp
+        return core.parallel_state.get_pipeline_model_parallel_rank() == total_current_pipeline_model_parallel_size - 1
```
aoyulong: Please add more comments for easier understanding.

flagscale/train/hetero/parallel_context.py
```diff
@@ -586,19 +595,25 @@ def _backtrack(mesh_index, prev_rank, path):
             ranks_list = current_mesh.get_all_process_group_ranks("pp")
             valid_ranks_list = []
             for ranks in ranks_list:
-                if prev_rank == -1 or (prev_rank, ranks[0]) in self._inter_mesh_process_groups:
+                mesh_is_connect = False
+                for prev_path_ranks in path:
+                    for prev_path_rank in prev_path_ranks:
+                        if (prev_path_rank, ranks[0]) in self._inter_mesh_process_groups:
+                            mesh_is_connect = True
```
aoyulong: Change mesh_is_connect to mesh_is_connected.

flagscale/train/hetero/parallel_context.py
Corle-hyz: It's a comment for fixing bug. The original `index()` method returns the position of the substring, for example, if the order is _"tp-cp-dp-pp"_ and the token is _"dp"_, then `order.index(token)` would return 6. However, we expect 2 to get its parallel size.

flagscale/train/hetero/p2p_communication.py
Corle-hyz: This is arbitrary because the shape of the recv tensor needs to be specified when communicating. It can be changed into any other shape.

flagscale/train/hetero/parallel_context.py
Corle-hyz: There is a conflict of ep between `validate_args` and `build_all_process_meshes`. I will add this argument back after supporting the validation of ep.

flagscale/train/hetero/parallel_context.py
Corle-hyz: OK.

flagscale/train/hetero/parallel_context.py
Corle-hyz: Each rank needs to store the `_inter_mesh_process_groups` information, so that during `_backtrack` we can judge whether two groups are inter-mesh connected.

megatron/megatron/core/pipeline_parallel/p2p_communication.py
Corle-hyz: OK.

flagscale/train/hetero/p2p_communication.py
Corle-hyz: OK.

flagscale/train/hetero/parallel_context.py
Corle-hyz: This argument is supported now.

aoyulong (approved):
LGTM

PR #198, opened 2024-08-29
[Hetero] Fix p2p comm in the hetero scenario

heavyrain-lzy:
1. Fix p2p comm in the hetero scenario because the commit https://github.com/NVIDIA/Megatron-LM/commit/8af3dae72a944848db0122047d89e04ab078b178
2. Change the `parallel_context` to support `ulysses sequence parallel`.

### DIFF
--- a/megatron/megatron/core/parallel_state.py
+++ b/megatron/megatron/core/parallel_state.py
@@ -1086,7 +1086,9 @@ def get_context_parallel_global_ranks(check_initialized=True):
 
 def get_ulysses_sp_parallel_group(check_initialized=True):
     """Get the ulysses sequence parallel group the caller rank belongs to."""
-    # TODO: support hetero
+    para_ctx = get_parallel_context() 
+    if para_ctx is not None:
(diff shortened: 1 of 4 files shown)


megatron/megatron/core/parallel_state.py
```diff
@@ -1847,7 +1855,7 @@ def get_data_modulo_expert_parallel_rank(with_context_parallel=False):
     """Return my rank for the context parallel group."""
     para_ctx = get_parallel_context() 
     if para_ctx is not None:
-        return para_ctx.get_data_modulo_expert_parallel_rank(with_context_parallel)
```
aoyulong, line 1850: Why does this code remove the `with_context_parallel` argument?

heavyrain-lzy: The `get_data_modulo_expert_parallel_rank` of `ParallelContext` doesn't support the argument  `with_context_parallel`.   Removing this argument is a temporary solution. The newest commit has added the `dp_modulo_exp` group to solve the problem.

aoyulong (approved):
LGTM

PR #199, opened 2024-08-29
[CI] Refactor the CI to be more modular and user-friendly

phoenixdong:
### Main features
- Only run the latest commit of the user's PR.
- Split unit tests by folder and functional tests by model into different jobs to avoid repeated runs when tests fail.
- Use YAML files to control the content of unit and functional tests.
- Before running the `pytest` command each time, check if the GPU is idle to avoid test failures caused by resource contention.

### How to run all unit tests locally
```bash
tests/scripts/unit_tests/test_all.sh
```

### How to run specific unit tests locally
```bash
tests/scripts/unit_tests/test_subset.sh --backend megatron --subset transformer
```

### How to add unit tests

- **Without adding a new folder:**
  - No modification needed.

- **When adding a new folder:**
  - Modify the `tests/scripts/unit_tests/config.yml` **configuration file**.
    
    **Configuration file structure**

    - **Framework name(megatron/flagscale/...):**
      - **set_environment:** Commands to configure the environment.
      - **coverage:** Directories that require test coverage statistics (the actual path = framework name + directory name).
  
    - **Subset name(models/transformer/...):**
      - **depth:** Specify which levels of tests to run. Default is to run all (`all`), also can be a integer.
      - **ignore:** Specify folders, files, or functions to skip.
      - **type:** Batch test or independent test. Default is batch (`batch`), also can be (`single`).

    **Note:** Default values do not need to be explicitly marked. If a folder uses all default configurations, it can be omitted from the configuration file.
    Here is an example of adding your own testing folder.
    ```
    flagscale:
      subset:
        your_folder:
          type: batch 
          depth: all 
          ignore: pytest_file.py
    ```
    How to run this example
    ```bash
    tests/scripts/unit_tests/test_subset.sh --backend flagscale --subset your_folder
    ```

### In the future, we will add unit testing and functional testing for VLLM


aoyulong (approved):
LGTM

PR #202, opened 2024-09-02
[Model] Update mixtral and llama config

Caozhou1995:
This PR adds mixtral tokenizer type and update llama config.


aoyulong (approved):
LGTM

PR #203, opened 2024-09-02
[Hetero] Fix hang in heterogeneous mode

heavyrain-lzy:
1. Fix hang in the heterogeneous mode, where there are more than one `proecc_meshse` and  the `tp_size` of one of them is `1` and the other is not `1`
2. Fix some other bugs


aoyulong (approved):
LGTM

PR #210, opened 2024-09-10
[BUG or ENHANCEMENT] Update qk_layernorm.

ftgreat:
With current qk_layernorm implement training did not converge. 
One shared qk_layernorm acts on every head, however qk_layernorm should affect all heads.
So just enlarge the shape of qk_layernorm weights, training converges as expected.

image

List some models using qk_layernorm for references:
1. https://github.com/mlfoundations/open_lm/blob/main/open_lm/model.py#L131 used in [DCLM-7B](https://huggingface.co/apple/DCLM-7B)
2. https://github.com/huggingface/transformers/blob/main/src/transformers/models/olmoe/modeling_olmoe.py#L394-L395 used in [OLMoE](https://huggingface.co/allenai/OLMoE-1B-7B-0924)

### DIFF
--- a/megatron/megatron/core/transformer/attention.py
+++ b/megatron/megatron/core/transformer/attention.py
@@ -360,7 +360,7 @@ class SelfAttention(Attention):
         if submodules.q_layernorm is not None:
             self.q_layernorm = build_module(
                 submodules.q_layernorm,
-                hidden_size=self.hidden_size_per_attention_head,
+                hidden_size=self.query_projection_size,
                 config=self.config,
                 eps=self.config.layernorm_epsilon,
             )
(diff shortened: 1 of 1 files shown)


megatron/megatron/core/transformer/attention.py
```diff
@@ -360,7 +360,7 @@ def __init__(
         if submodules.q_layernorm is not None:
             self.q_layernorm = build_module(
                 submodules.q_layernorm,
-                hidden_size=self.hidden_size_per_attention_head,
+                hidden_size=self.query_projection_size,
```
heavyrain-lzy, line 363: Thanks for your contribution.  According to my understanding, the difference between the origin code and your code is as follows:

-  origin code: normalize the every single head individually
-  new code: normalize all the heads together

According to the `input_layernorm`https://github.com/NVIDIA/Megatron-LM/blob/bbecd0812ffc9a90eef472fda91c60eead3f417f/megatron/core/transformer/transformer_layer.py#L111, we should normalize all the heads together. Maybe you can add an issue in `Megatron-LM` to recheck it.
I think you are right.
LGTM.

PR #224, opened 2024-09-25
[FlagScale] Split args associated with process_mesh

heavyrain-lzy:
Update the `args` and `xxx_config` according to  the associated process_mesh.


flagscale/train/arguments.py
```diff
@@ -0,0 +1,228 @@
...
+            return self._rank_mapper
+        return getattr(self.args, name)
+
+    def __str__(self):
+        return str(self.args)
+
+    def __repr__(self):
+        return repr(self.args)
+    
+    def _initialize_distributed(self):
```
aoyulong, line 27: Why do we need to move this distributed initialization here?

flagscale/train/arguments.py
```diff
@@ -0,0 +1,228 @@
...
+    def _build_rank_mapper(self):
+        self._initialize_distributed()
+        self._rank_mapper = RankMapper(self.args)
+        return self._rank_mapper
+
+    @classmethod
+    def from_dict(cls, args):
+        return cls(args)
+
+    def pre_validate_args(self):
```
aoyulong, line 69: The arguments extended by flagscale were split into two parts for now. However, in the future, the arguments may have more than two parts.

flagscale/train/arguments.py
heavyrain-lzy: To create the map between `logical_rank` and `physical rank`.

flagscale/train/arguments.py
heavyrain-lzy: To decrease the code change in the `Megatron-LM` repo, limit modifications to the `validate_args` function in the original argument file when adding new parameters.

megatron/megatron/training/initialize.py
```diff
@@ -274,7 +282,7 @@ def _initialize_distributed(get_embedding_ranks, get_position_embedding_ranks):
     # data-parallel communicators.
     if device_count > 0:
         # Set the parallel context.
-        if args.enable_hetero:
```
aoyulong: So we can remove the `args.enbale_hetero` from the code and use `args.hetero_process_meshes` instead?

aoyulong (commented):
We should design a better mechanism to extend new args by FlagScale.

megatron/megatron/training/initialize.py
heavyrain-lzy: I think that the way using `process_meshes` to configure the training plan is not only designed for `heterogeneous training`, but also for other multi-model training like `Llava`. So we should unbind the `args.enbale_hetero` parameter and the `hetero_process_meshes` parameter gradually.

aoyulong (approved):
LGTM

PR #228, opened 2024-10-05
[metax] add mixtral-8x7b flagscale continue train patch

lytm00847:
[metax] add mixtral-8x7b flagscale continue train patch


aoyulong (approved):
LGTM

PR #246, opened 2024-10-29
[Hetero] Optimize cpu communication

heavyrain-lzy:
1. The p2p in ProcessMesh uses the `nccl` group to communicate.
2. Add an argument to decide whether to use CPU communication.
3. Change `get_last_rank_when_using_pipeline` according to the function in `parallel_state`


aoyulong (approved):
LGTM

PR #298, opened 2024-12-26
[Serve] multiple submodels serve

cyber-pioneer:
## Type
New Feature

## Description

1. **multiple submodels serve** 
Given many submodules. Simply configure the keywords of the submodules, including resource allocation and interdependencies among the submodels.
2. **multiple nodes serve** 
Given many  well-connected nodes. Simply configure the keywords of the nodes, including resource type and address messages.
3. **ci support serve case** 
Serve test case contains 2 steps: serve and call.


flagscale/runner/runner_serve.py
```diff
@@ -100,17 +100,14 @@ def _generate_run_script_serve(
         f.write(f"\n")
         f.write(f'cmd="{cmd}"\n')
         f.write(f"\n")
-        if with_test:
-            f.write(f'bash -c "$cmd; sync" \n')
+        # TODO: need a option to control whether to append or overwrite the output file
+        # Now, it always appends to the output file
```
aoyulong, line 104: Yes, if there is an option available, users will be satisfied.

flagscale/runner/runner_serve.py
```diff
@@ -162,10 +160,17 @@ def _prepare(self):
         _update_config_serve(self.config)
         self.user_args = _get_args_vllm(self.config)
         self.user_envs = self.config.experiment.get("envs", {})
+        entrypoint = self.config.experiment.task.get("entrypoint")
         if self.command_line_mode:
             self.user_script = "flagscale/serve/run_vllm.py"
+        elif isinstance(entrypoint, str) and entrypoint.endswith(".py"):
+            self.user_script = entrypoint
+        elif entrypoint is None:
```
aoyulong, line 168: Should we require users to provide the entrypoint?

flagscale/serve/core/dag.py
```diff
@@ -0,0 +1,455 @@
...
+from pydantic import create_model
+from typing import Callable, Any
+from fastapi import FastAPI, HTTPException, Request
+import ray
+from ray import workflow
+
+from flagscale.logger import logger
+
+
+class Builder:
```
aoyulong: Builder for what? I believe the name should be more specific.

flagscale/serve/core/dag.py
```diff
@@ -0,0 +1,455 @@
...
+        return list(output_node)[0]
+
+    def check_and_get_port(self, target_port=None, host="0.0.0.0"):
+        """
+        Check if a specific port is free; if not, allocate a free port.
+        :param target_port: The port number to check, default is None.
+        :param host: The host address to check, default is "0.0.0.0".
+        :return: A tuple (is_free, port), where `is_free` indicates if the target port is free,
+                and `port` is the allocated port (target_port if free, or a new free port).
+        """
```
aoyulong: The code should be placed in the utilities so that it can be reused.

flagscale/runner/runner_serve.py
cyber-pioneer: This is an option. Models pipeline can be built via user's entrypoint file or yaml config.

flagscale/serve/core/dag.py
cyber-pioneer: ok

flagscale/serve/core/dag.py
cyber-pioneer: done

aoyulong (approved):
LGTM

PR #300, opened 2024-12-30
Update runner_train.py, “no_shared_fs" should be under the field "runner"

changsci:
“no_shared_fs" should be under the field "runner"

Change
no_shared_fs = config.experiment.get("no_shared_fs", False) 
to 
no_shared_fs = runner_config.get("no_shared_fs", False)


aoyulong:
Thank you for your valuable input! Your contribution is greatly appreciated.

PR #310, opened 2025-01-07
[Core] Support sigmoid router for aux_loss in moe

lxd-cumt:
Support sigmoid router in moe similar to DeepSeek V3
- add a configuration str <moe_router_score_function_type> in TransformerConfig
- support using sigmoid+normalization to compute prob scores

The implementation in this PR, is similar to that in DeepSeek V3, as shown in the figure below. Scores are first calculated using a sigmoid for x, and then the topk function is used to select the k largest scores, finally the k scores are averaged.

Currently, sigmoid router is supported for two load balancing types: auxiliary loss and sequence auxiliary loss.
We will support auxiliary loss free load balancing in the future.

references to: https://github.com/deepseek-ai/DeepSeek-V3/blob/main/DeepSeek_V3.pdf

### DIFF
--- a/megatron/megatron/core/transformer/moe/moe_utils.py
+++ b/megatron/megatron/core/transformer/moe/moe_utils.py
@@ -94,6 +94,17 @@ def sequence_load_balancing_loss_func(
 
     return seq_aux_loss
 
+def score_function(
+    input: torch.Tensor,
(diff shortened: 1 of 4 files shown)


aoyulong (approved):
LGTM

PR #315, opened 2025-01-17
[Core] Support to skip samples according to spiky loss

heavyrain-lzy:
Support to skip samples according to spiky loss. There are two methods to skip samples:
1.    You can manually set the skipping range
```
skip_samples_range: [20, 30]
```
2.    Automatically detect the spiky loss and skip the samples
```
auto_skip_spiky_loss: true
spiky_loss_threshold: 0.25
```

### DIFF
--- a/flagscale/train/train.py
+++ b/flagscale/train/train.py
@@ -97,7 +97,7 @@ from megatron.training import ft_integration
 from flagscale.train.extra_valid import extra_evaluate_and_print_results
 from flagscale.train.extra_valid import build_extra_valid_data_iterators
 from flagscale.train.stablelm2_scheduler import StableLM2SchedulerConfig
-from flagscale.train.global_vars import get_parallel_context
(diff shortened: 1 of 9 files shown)


flagscale/train/train.py
```diff
@@ -832,6 +832,18 @@ def train_step(forward_step_func, data_iterator,
     if should_exit:
         return {}, True, should_checkpoint, should_exit, exit_code, None, None
 
+    ###########FlagScale begin ##########
```
aoyulong: We need to define the exact format for commenting on the beginning and the ending of the customized code from FlagScale:
* ` ########## FlagScale Begin ##########`: 10*`#` + `[Space]FlagScale Begin[Space]` + 10*`#`
* ` ########## FlagScale End ##########`: 10*`#` + `[Space]FlagScale End[Space]`+ 10*`#`

heavyrain-lzy: Done. Thanks!

aoyulong (approved):
LGTM

PR #321, opened 2025-02-10
[Core] Update inference/core for vllm-v0.7.1

zhaoyinglia:
- update inference/core for vllm-v0.7.1
- fix the yaml management function with omegaconf
- fix the automatic identification function of the vllm installation path


aoyulong (approved):
LGTM

PR #325, opened 2025-02-11
[Serve] Make hostfile config the same as train and inference

cyber-pioneer:
## Type
Improvement

## Description

Make hostfile config the same as train and inference:
If multiple nodes are needed in serve, hostfile should be provided.
Example:
```
# ip slots type=xxx[optional]
# master node
x.x.x.x slots=8 type=gpu 
# worker nodes
x.x.x.x slots=8 type=gpu
```


aoyulong (commented):
Let’s do our best to keep the host file simple by avoiding any extra fields. This will help us stay organized and make things easier for everyone!

cyber-pioneer:
> Let’s do our best to keep the host file simple by avoiding any extra fields. This will help us stay organized and make things easier for everyone!

right!

aoyulong (approved):
LGTM

PR #337, opened 2025-02-14
[Model][Experimental]Support multi token predictor

lxd-cumt:
Support multi token predictor used in Deepseek V3 pretrain.

- support adding N multi token predictors behind the model to participate in training;
- support tensor parallelism, pipeline parallelism, expert parallelism for mtp. Where using pipeline parallelism, the MTP module is added to the last PP stage;
- support checkpoint convert for mtp modules.

next pr: #340

### DIFF
--- a/flagscale/train/models/deepseek_v3/multi_token_predictor.py
+++ b/flagscale/train/models/deepseek_v3/multi_token_predictor.py
@@ -54,15 +54,15 @@ class DeepSeekSharedEmbedding(MegatronModule):
             scatter_to_sequence_parallel=scatter_to_sequence_parallel,
         )
         self.embedding.word_embeddings.weight.is_embedding_or_output_parameter = True
-        
-    
+
     def forward(
         self,
         input_ids,
         position_ids,
     ) -> Tensor:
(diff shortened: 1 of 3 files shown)


flagscale/train/models/deepseek_v3/multi_token_predictor.py
```diff
@@ -0,0 +1,293 @@
...
+            logits_mtp, pre_hidden_states = self.mtp_modules[i](
+                input_ids=input_ids,
+                position_ids=position_ids,
+                attention_mask=attention_mask,
+                pre_hidden_states=pre_hidden_states,
+                decoder_input=decoder_input,
+                input_mask=input_mask,
+            )
+            logits_mtps.append(logits_mtp.transpose(0, 1).contiguous())
+            input_mask = roll_input_mask(input_mask)
```
heavyrain-lzy, line 290: As shown below:
image
The input sequences should be `[t1,t2,t3,t4], [t2,t3,t4,t5]` not `[t1,t2,t3,t4, t5], [t2,t3,t4,t5]` and `attention_mask` should keep consistent with `input_ids`

PR #338, opened 2025-02-17
[Operator] Add USE_FLAGGEMS Option

kiddyjinjin:
## description
This update introduces conditional support for` flag_gems `based on the `USE_FLAGGEMMS `environment variable. After both the async and sync engines are initialized, the code attempts to import and enable `flag_gems` as the default ops implementation. If the import fails, it logs a warning and falls back to the default implementation.


aoyulong (dismissed):
Do we need a method to assist users in selecting which operator should be replaced?

vllm/vllm/engine/async_llm_engine.py
```diff
@@ -650,6 +650,17 @@ def from_engine_args(
             usage_context=usage_context,
             stat_loggers=stat_loggers,
         )
+
+        if os.getenv("USE_FLAGGEMMS", "false").lower() in ("1", "true", "yes"):
```
aoyulong: It would be beneficial to include a link to FlagGems for users.

0x45f: USE_FLAGGEMMS --> USE_FLAGGEMS？

PR #340, opened 2025-02-17
[Model][Experimental] Support share embedding, sp/cp parallel for MTP

lxd-cumt:
Support more features for deepseek v3 multi token predictor , refs pr: #337 

- [x] - support shared embedding/output lm head between mtp and main model
- [x] - support sequence parallelism, context parallelism
- [x] - support loss computation when passing labels to DeepSeekV3Model
- [x] - fix bugs in rolling input of mtp
- [x] - remove redundant config: use_mtp_predictor
- [ ] -  init mtp embedding only when using pipeline parallel, reuse main model embedding when not using pp
- [ ] -  simultaneously support share main model embedding with main model output lm head, and main model embedding with mtp embedding
- [ ] -  use layer spec


flagscale/train/models/deepseek_v3/deepseek_v3_model.py
```diff
@@ -103,23 +108,32 @@ def __init__(
             scatter_embedding_sequence_parallel=scatter_embedding_sequence_parallel,
             seq_len_interpolation_factor=seq_len_interpolation_factor,
         )
-        
-        if self.post_process and self.use_mtp_predictor:
+
+        if self.use_mtp_predictor and self.post_process:
```
heavyrain-lzy, line 112: If pipeline parallelism is used, the `MTP` module is in the last pipeline stage, so the `self.post_process` is always `True.`

flagscale/train/models/deepseek_v3/deepseek_v3_model.py
```diff
@@ -205,44 +219,155 @@ def forward(
         if not self.post_process:
             return hidden_states
 
-        # logits and loss
         # logits of main model
         output_weight = None
         if self.share_embeddings_and_output_weights:
             output_weight = self.shared_embedding_or_output_weight()
         logits, _ = self.output_layer(
```
heavyrain-lzy, line 226: How to share `output_layer` with `main_model`?

zhaoyinglia (approved):
LGTM

PR #348, opened 2025-02-19
[Serve] serve supports other device more than gpu and cpu

cyber-pioneer:
## Type
Improvement

## Description

in config.yaml, specify **nproc_per_node** and **device_type** on target device.
Note: config key  **device_type** is specified for _**single node serving**_ in neither gpu or cpu. In **_multiple nodes_** case, use config key **hostfile** instead.
```
experiment:
  runner:
    nproc_per_node: 8
    device_type: xxx
```


aoyulong (approved):
LGMT

PR #351, opened 2025-02-20
Support MTP layer spec, and both vertical and horizontal parameter sharing

lxd-cumt:
Support more mtp features

- [ ] -  init mtp embedding only when using pipeline parallel, reuse main model embedding when not using pp
- [x] -  simultaneously support share main model embedding with main model output lm head (vertical), and main model embedding with mtp embedding (horizontal)
- [x] -  use layer spec
- [x] - fix rope fusion bugs


zhaoyinglia (approved):
LGTM

PR #361, opened 2025-02-21
[Core] Move Autotuner to Runner

zhaoyinglia:


aoyulong (approved):
LGTM. We should add some unit tests, and it's better to create an AutoTuner class, along with separate TrainAutoTuner and ServeAutoTuner classes.

PR #363, opened 2025-02-23
[Serve] Support serve via origin engine

cyber-pioneer:
## Type
Feature

## Description

Support serve via particular backend like vllm by specify **use_native_serve**
```
experiment:
  task:
    inference_engine: vllm

serve:
  deploy:
    use_native_serve: False
```

### DIFF
--- a/examples/deepseek/conf/config_deepseek_v3.yaml
+++ b/examples/deepseek/conf/config_deepseek_v3.yaml
@@ -10,7 +10,7 @@ experiment:
     backend: vllm
     entrypoint: null
   runner:
-    hostfile: examples/deepseek/conf/hostfile.txt # absolute path to hostfile
+    hostfile: examples/deepseek/conf/hostfile.txt
     docker: ds
   envs:
     CUDA_DEVICE_MAX_CONNECTIONS: 1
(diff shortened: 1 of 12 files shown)


aoyulong (commented):
It's better to change the config to `use_native_serving: False`

examples/deepseek/conf/config_deepseek_v3.yaml
```diff
@@ -10,7 +10,7 @@ experiment:
     backend: vllm
     entrypoint: null
   runner:
-    hostfile: examples/deepseek/conf/hostfile.txt
```
aoyulong: I think it's appropriate to allow users to provide a relative host file path in the current directory, which we can then convert to absolute paths.

flagscale/serve/run_backend.py
```diff
@@ -0,0 +1,46 @@
...
+    stdout, stderr = process.communicate()
+    logger.info(f"[Serve]: Standard Output: {stdout}")
+    logger.info(f"[Serve]: Standard Error: {stderr}")
+
+    return process.returncode
+
+
+def main():
+    serve.load_args()
+    backend = serve.task_config.experiment.task.get("backend", None)
```
aoyulong: It's better to change `backend` to `inference_engine` since we use `backend` in other places.

cyber-pioneer:
> It's better to change the config to `use_native_serving: False`

done

flagscale/serve/run_backend.py
cyber-pioneer: done

aoyulong (approved):
LGTM

PR #371, opened 2025-02-24
[Core] Fix ucc comm-backend when using `FlagCX` backend

heavyrain-lzy:
Change the default value of  `pipeline-model-parallel-comm-backend` to `None` to support the `distributed-backend` is not `nccl`.


aoyulong (approved):
LGTM

PR #381, opened 2025-02-27
[CLI] Add requirements and improve ease of use

Caozhou1995:
This PR sets the requirements of the cli installation, and we improve ease of use:

- flagscale pull just on host machine, so we moved some imports into concrete functions

- To avoid installation error on the CPU without GPU, we fix the exception


aoyulong (approved):
LGTM

PR #384, opened 2025-02-28
[CherryPick][Serve] Fix the case of deepseek r1 (#383)

cyber-pioneer:
### Description
Fix example case deepseek r1.


aoyulong (approved):
LGTM

PR #386, opened 2025-02-28
[Core] Fix import error

Caozhou1995:
To accommodate the scenario where the before_start field is used to switch to the actual environment during program execution,  we have placed the import statements inside the function body rather than at the beginning of the file.


aoyulong (approved):
LGTM

PR #389, opened 2025-02-28
a preliminary implementation of single-process simulator, which still has problems

Corle-hyz:
Support to simulate a single rank of distributed training. Rank 0 is allowed now. Also, this PR includes a preliminary implementation of auto-tuning in cooperation with the simulator.

INFO: This simulator will not normally exit, so it must be killed manually. Further improvement is needed.


tengqm:
@Corle-hyz Are you still working on this PR? Please let the team know if helps are needed. It has been a long time since the last commit. Or we can close this PR and reopen it when you got cycles to push this forward?

PR #392, opened 2025-03-02
[CI] Update some installation details

phoenixdong:
1. Using the default source of apt.
2. Add some packages that serve needs.
3. Split some major installation steps.


aoyulong (approved):
LGTM

PR #408, opened 2025-03-13
[Core] Add a general estimator

aoyulong:
This PR introduces a new estimator folder with modules for estimating model resource usage. It provides:

- Theoriatical estimation for parameters, FLOPs, and activations.
- Utilities for computing memory requirements and printing results.
- Configurations and modules (attention, MLP, transformer) to simulate large models (e.g., GPT).
- These additions help analyze memory and compute demands during model design and scaling.
- Support data parallelism, tensor parallelism ,and pipeline parallelism.

Users can use this tool on their PCs without the need for a GPU.
```
cd FlagScale/tools
python estimate_gpt.py
```

### DIFF
--- a/flagscale/runner/estimator/utils.py
+++ b/flagscale/runner/estimator/utils.py
@@ -1,0 +1,142 @@
+from typing import Any, Dict
+from flagscale.runner.estimator.meta_base import ModelConfig, register_model, get_registry
+
+def compute_memory(config: ModelConfig, params: int, acts: int) -> int:
+    """
+    Compute memory requirements for the model.
+    
+    Parameters:
+    -----------
+    config : ModelConfig
+        Model configuration
+    params : int
+        Number of parameters
+    acts : int
+        Number of activations
+        
+    Returns:
+    --------
+    int
+        Memory requirement in bytes
+    """
+    # Determine parameter bytes based on dtype
+    dtype_bytes = {
+        "fp32": 4,
+        "fp16": 2,
+        "bf16": 2,
+        "float32": 4,
+        "float16": 2,
+        "bfloat16": 2,
+    }
+    
+    # Use bf16 as default if dtype is not specified or not recognized
+    param_dtype = getattr(config, "dtype", "bf16")
+    param_bytes = dtype_bytes.get(param_dtype, 2)
+    
+    # Check if distributed optimizer is enabled
+    use_distributed_optimizer = getattr(config, "use_distributed_optimizer", False)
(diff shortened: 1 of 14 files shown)


flagscale/runner/estimator/utils.py
```diff
@@ -0,0 +1,92 @@
...
+    param_bytes = 2
+    
+    # Static memory for parameters, optimizer states, and gradients
+    # - Parameters: 2 bytes per param (bf16)
+    # - Optimizer states: 8 bytes per param (e.g., Adam with m and v states, each is 4 bytes)
+    # - Gradients: 2 bytes per param (bf16)
+    param_memory = device_params * (param_bytes + 8 + param_bytes)
+    
+    # Dynamic memory for activations (2 bytes per activation)
+    act_memory = acts * 2
```
Corle-hyz, line 35: Enabling `use_distributed_optimizer` is a common practice during training and is also the default configuration in FlagScale. Should we also evaluate its impact on optimizer memory usage?

flagscale/runner/estimator/utils.py
```diff
@@ -0,0 +1,92 @@
...
+    device_params = params / (config.tensor_parallel_size * config.pipeline_parallel_size)
+    
+    # Assume bf16 precision for training (2 bytes per parameter)
+    param_bytes = 2
+    
+    # Static memory for parameters, optimizer states, and gradients
+    # - Parameters: 2 bytes per param (bf16)
+    # - Optimizer states: 8 bytes per param (e.g., Adam with m and v states, each is 4 bytes)
+    # - Gradients: 2 bytes per param (bf16)
+    param_memory = device_params * (param_bytes + 8 + param_bytes)
```
Corle-hyz: Enabling `use_distributed_optimizer` is a common practice during training and is also the default configuration in FlagScale. Should we also evaluate its impact on optimizer memory usage?

Corle-hyz: Given that the embedding layer is tied to the first pipeline parallel (PP) stage, should we exclude it from the total parameter count when assessing the impact of pipeline parallelism?

Corle-hyz: The `accumulate_allreduce_grads_in_fp32` parameter specifies that a main gradient should be stored in fp32 for accumulation and all-reduce operations. This suggests that the optimizer may require an additional 4 bytes of memory per gradient.

zhaoyinglia (approved):
LGTM

PR #412, opened 2025-03-14
Fix expert_bias saving and loading bugs in moe router

lxd-cumt:
- remove redundant score_bias, and update ckpt conversion by loading data to "module.expert_bias"
- fix ckpt conversion bugs of deepseek v3 when using even pipeline parallel


zhaoyinglia (approved):
LGTM

PR #395, opened 2025-03-04
[Core] Support the MoE heterogenous training of DeepSeek V3 

heavyrain-lzy:
Support hetero-ep in different process-meshes to train the models based on `MOE` architecture. This PR also supports pretraining `deepseek-v3` model from scratch.
There are some limits when using hetero-ep temporarily:
1. All ep-size in process-meshes should be 1 or none is 1!
2. When `num_mtp_predictor` is not 0, the `tp-size` between `first-pp-stage` and `last-pp-stage` must be the same.

Some tests using `examples/deepseek_v3/conf/train/train_deepseek_v3_hetero.yaml` are as flows:
enbale_hetero | parallism | pp_layer | etp | num_mtp_predictor | grad_norm of first step | 10step loss | rank0-memory(max-reserved)
-- | -- | -- | -- | -- | -- | -- | --
No | pp2dp2ep2tp2 | 12,12 | 1 | 1 | 14.081 | 11.64818 | 52178
No | pp2dp2ep2tp2 | 12,12 | 2 | 1 | 13.682 | 11.65728 | 37814
Yes | dp4ep2-dp4ep4 | 13,11 | 1 | 1 | 14.727 | 11.66628 | 58302
Yes | dp4ep4-dp2ep2-dp2ep2 | 11, 6, 7 | 1 | 1 | 14.025 | 11.65973 | 61422

PR #429, opened 2025-03-26
[Fix] Fix hetero etp & update aquial_3B config

heavyrain-lzy:
1. Fix the bug that we can't use `expert_tensor_parallel_size_per_process_mesh` to set `expert_tensor_parallel_size` in hetero mode in https://github.com/FlagOpen/FlagScale/pull/395
2. Update the `Aquila-3B` configuration.


aoyulong (approved):
LGTM

PR #432, opened 2025-03-27
[Core] Improve cpu-based communication to support MoE models

heavyrain-lzy:
1. Improve CPU-based communication to support MoE models.
2. Improve CPU-based communication to support models with `shared_embedding` and `MTP`


aoyulong (approved):
LGTM

PR #438, opened 2025-04-01
[BugFix] Fix patch bug

Caozhou1995:
This PR fixes a bug in the patch mechanism, preventing the deletion of other patches after generating a new patch.


aoyulong (approved):
LGTM

PR #445, opened 2025-04-09
[Feature]support tp for deepseek checkpoint conversion

lxd-cumt:
support tp for deepseek checkpoint conversion

### DIFF
--- a/tools/checkpoint/deepseek_v3/ckpt.py
+++ b/tools/checkpoint/deepseek_v3/ckpt.py
@@ -98,7 +98,6 @@ def get_hf_mtp_ckpt(message, model, mtp_layer_id, args):
 #### set to megatron ckpt
 def set_embedding_ckpt(message, models, md, args):
     tp_size, _, _, _ = _get_parallel_size(args)
-    # assert tp_size == 1, "do not support TP parallel for deepseek v3 currently"
 
     # embedding
     pos_embed = None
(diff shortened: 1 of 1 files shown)


tools/checkpoint/deepseek_v3/ckpt.py
```diff
@@ -459,74 +496,82 @@ def get_moe_mlp_ckpt(message, models, layer_id, args):
...
-                    expert_down_weight = expert.lienar_fc2.weight.data
+                    expert_linear1_weight.append(expert.linear_fc1.weight.data)
+                    expert_linear2_weight.append(expert.linear_fc2.weight.data)
                 else:  # using TEGroupedMLP
-                    expert_linear_fc1_weight = getattr(
-                        tf_layer.mlp.experts.linear_fc1, f"weight{expert_id}", None
+                    expert_linear1_weight.append(
+                        getattr(
+                            tf_layer.mlp.experts.linear_fc1, f"weight{expert_id}", None
+                        ).detach()
```
zhaoyinglia, line 515: why detach?

lxd-cumt: The function **getattr()** return a non-leaf tensor with grad func, but is does support corss-process boundaries.
Therefore, if we want to transfer data by using multi-process queue in ckpt conversion, we need to call detach() on the tensor before serializing.

zhaoyinglia (approved):
LGTM

PR #457, opened 2025-04-18
[Serve] Support xPyD Disaggregation in multiple nodes

cyber-pioneer:
### Description: Multi-node Prefill/Decode Disaggregated Deployment with FlagCX

This PR implements support for multi-node disaggregated deployment of **prefill** and **decode** stages using `xPyD` Disaggregation:
- Schedule strategies of PD instances currently support: `robin`, `random`. default is `robin`.
- It introduces a new communication backend based on [FlagCX](https://github.com/FlagOpen/FlagCX). Merge [FlagCX Adapter](https://github.com/FlagOpen/FlagScale/pull/461).
- KV cache transfer is enabled via [p2pConnector](https://github.com/vllm-project/vllm/pull/15806) in `vLLM`.

---

### How to Use

**Step 1**: Install [FlagCX](https://github.com/FlagOpen/FlagCX?tab=readme-ov-file#quick-start)

**Step 2**: Install the `vLLM` version from [FlagScale](https://github.com/FlagOpen/FlagScale?tab=readme-ov-file#setup)

**Step 3**: Define your config files under `./examples/qwen/conf`

**Step 4**: Launch the distributed deployment  
```bash
python run.py --config-path ./examples/qwen/conf --config-name config_qwen2.5_7b_disagg_xpyd action=run
```

**Step 5**: Send requests to the deployed service  
```bash
curl -X POST -s http://localhost:10001/v1/completions \
-H "Content-Type: application/json" \
-d '{
  "model": "/models/Qwen2.5-7B-Instruct",
  "prompt": "Introduce Bruce Lee in details",
  "max_tokens": 100,
  "temperature": 0,
  "stream": true
}'
```


third_party/vllm
aoyulong, line 1: Will this change influence our vllm patch?

flagscale/serve/run_pd_disagg_router.py
aoyulong, line 1: Can we come up with a suitable abbreviation for "prefill-decoding disaggregation"?

flagscale/runner/runner_serve.py
```diff
@@ -22,6 +24,141 @@
...
+    allocated_ports = []
+    for i in range(num):
+        port = get_free_port()
+        while port in allocated_ports or port in exclude_ports:
+            port = get_free_port()
+        allocated_ports.append(port)
+    return allocated_ports
+
+
+class ResourceManager:
```
aoyulong: ResourceManager can be utilized not only for serving but also for other components. Therefore, it can be extracted for future reuse.

third_party/vllm
cyber-pioneer: This pr adds xPyD-relative files in vllm patch. And current function would not be influenced.

flagscale/serve/run_pd_disagg_router.py
cyber-pioneer: PD-Disaggregation seems common. Renamed as pd_disaggregation.

flagscale/runner/runner_serve.py
cyber-pioneer: Nice idea. Done

aoyulong (approved):
LGTM

PR #462, opened 2025-04-18
[Comm] Add a communicator for the edge-cloud collaborative

phoenixdong:
**Edge-cloud collaborative communicator to support robot agent system:**

1. Create an agent registration interface.
2. Implement a command sending interface.
3. Implement a command listening interface that is blocking rather than polling, offering better performance.
4. Develop data recording and access interfaces that support both partial and full access.

**Features:**

1. Utilizes Redis's publish/subscribe model at the lower level to achieve high-performance command publishing and responding, with response times reaching under 0.001 seconds.
2. Supports bidirectional communication between the brain and the robot, facilitating a closed-loop control path between brain commands and robot feedback.

**Usage:**

1. Include the source code.
2. Run: `pip install git+https://github.com/FlagOpen/FlagScale.git`, import with: `from flag_scale.communication import Communicator`

### DIFF
--- a/communication/__init__.py
+++ b/communication/__init__.py
@@ -1,0 +1,1 @@
+from .communicator import Communicator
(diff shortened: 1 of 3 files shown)


aoyulong (approved):
We should improve our file organization in the next pull request.

PR #466, opened 2025-04-22
[BugFix] Fix the cli bug for submodule

Caozhou1995:
This PR fixes the CLI issue, where the backend is now managed as a submodule, but previously it was still packaged as a subtree.


aoyulong (approved):
LGTM

PR #469, opened 2025-04-23
[Tools] Speed up the llavaov data preprocess

Caozhou1995:
This PR accelerates the data preprocessing of Llava OneVision. It adopts a multi-task approach, launching multiple tasks simultaneously for processing, with each task handling a portion of the data. In practical tests, it can achieve an acceleration of more than twice in the single image scenario, and an acceleration of up to 1.4 times in the video and multi-image scenarios.
Usage: set task count by `twice` in  `make_llava_ov_wds.sh`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,10 +74,11 @@ setup(
     },
     install_requires=[
         "click",
+        "gitpython",
         "cryptography",
         "setuptools>=75.1.0",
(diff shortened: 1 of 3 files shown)


aoyulong (approved):
LGTM

aoyulong (approved):
LGTM

PR #488, opened 2025-04-28
[BugFix] Fix bug when using `torch.load` of `pytorch>=2.6`

heavyrain-lzy:
Fix bug when using torch.load of `pytorch>=2.6`. We've only modified `torch.load` as follows:
1. The relevant files that we use for training right now
2. The files that swallowed the error in `torch.load`.

This way, we can minimize the intrusion of modifying Megatron-related files while keeping them from being misused.


aoyulong (approved):
LGTM

PR #510, opened 2025-05-08
Add Llama.cpp Inference Backend

Hchnr:
Add llama.cpp as a new inference backend, changes:
1. Patch&unpatch support backend "llama.cpp"
2. Add new submodule llama.cpp
3. Call llama-server(build output) in run_inference_engine.py
4. Add 3 script to build llama.cpp and convert checkpoint
5. Add an example conf


ceci3 (dismissed):
LGTM

.gitmodules
```diff
@@ -10,3 +10,6 @@
 	path = third_party/Megatron-Energon
 	url = https://github.com/NVIDIA/Megatron-Energon.git
 	branch = develop
+[submodule "third_party/llama.cpp"]
+	path = third_party/llama.cpp
+	url = git@github.com:ggml-org/llama.cpp.git
```
aoyulong: Please use HTTP like others do.

flagscale/backends/llama.cpp/perf.md
```diff
@@ -0,0 +1,35 @@
...
+Model: https://huggingface.co/Qwen/Qwen3-0.6B
+
+## Apple M4 (10 core)
+
+### CPU
+
+| model                          |       size |     params | backend    | threads |            test |                  t/s |
+| ------------------------------ | ---------: | ---------: | ---------- | ------: | --------------: | -------------------: |
+| qwen3 0.6B F16                 |   1.40 GiB |   751.63 M | Metal,BLAS |       4 |           pp512 |        772.66 ± 9.36 |
+| qwen3 0.6B F16                 |   1.40 GiB |   751.63 M | Metal,BLAS |       4 |           tg128 |         54.98 ± 2.62 |
```
aoyulong: What do pp512 and tg128 mean? A brief explanation would be helpful.

flagscale/serve/run_inference_engine.py
```diff
@@ -45,7 +69,7 @@ def main():
 
     model_config = None
     for item in serve_config:
-        if item.get("serve_id", None) == "vllm_model":
+        if item.get("serve_id", None) in ["vllm_model", "llama_cpp_model"]:
```
aoyulong: I believe the serve_id distinguishes the current serving from others. So, why do we only allow vllm_model or llama_cpp_model?

examples/qwen3/conf/serve/0_6b_llama_cpp.yaml
```diff
@@ -0,0 +1,13 @@
+- serve_id: llama_cpp_model
+  engine: llama_cpp
+  engine_args:
+    model: /tmp/models/Qwen3-0.6B/ggml_model_f16.gguf
```
aoyulong: Since there is no argument mapping between vllm and llama.cpp, do they share the same arguments?

flagscale/backends/llama.cpp/flagscale_build.sh
aoyulong, line 1: These shell scripts may need to be relocated to different locations. Besides, why do we need three scripts?

examples/qwen3/conf/serve/0_6b_llama_cpp.yaml
Hchnr: Arguments below are fake for now, I should remove them.

flagscale/backends/llama.cpp/flagscale_build.sh
Hchnr: Merge 3 scripts in 1 LGTM, 3 scripts is for users to do building and converting seperately. 
Locate the scripts here is for keeping same with the other backends (building at third_party/xxx).

Hchnr:
Updates:
1. Update README.md
  1.1. add converting (hf->gguf) tools example
  1.2. add some illustration for perf report
2. Add args mapping from vllm to llama.cpp, as the table below
3. Update  install/install-requirement.sh
  3.1. Do installing of llama.cpp here
  3.2. Adjust order, fix some invalid path
image

flagscale/serve/args_mapping/llama_cpp.py
```diff
@@ -0,0 +1,58 @@
...
+# Serve args(keys and values) mapping from vllm to llama_cpp
+VLLM_LLAMA_KV_MAPPING = {
+    "uvicorn_log_level": kv_mapping_uvicorn_log_level,
+    "lora_modules": kv_mapping_lora_modules,
+    "max_model_len": kv_mapping_max_model_len,
+    "reasoning_parser": kv_mapping_reasoning_parser,
+    "kv_cache_dtype": kv_mapping_kv_cache_dtype,
+    "speculative_config": kv_mapping_speculative_config,
+    "override_pooler_config": kv_mapping_override_pooler_config,
+}
```
aoyulong: Could we move these mappings into a YAML file for better management?

Caozhou1995 (commented):
LGTM

ceci3 (approved):
LGTM

PR #515, opened 2025-05-10
[Tools] Add patch encryption and enhance patch usability

Caozhou1995:
This PR adds patch encryption, which can be used with the command `python tools/patch/patch.py --key-path`, and for decryption with` python tools/patch/unpatch.py --key-path.` In addition, this PR further improves the usability of the patch mechanism, including: 

- Support for fuzzy matching of backend inputs, for example, Megatron-LM can be entered as megatron or megatron-lm.
- The patch generation process has been enhanced to support modifications involving subtrees.
- When the vendor performs `unpatch`, if neither `patch_history.yaml` is present nor the `commit` parameter is provided, the patch file from the current commit will be used for unpatching by default.
- Automatic check for Git config has been added.


aoyulong (approved):
LGTM

PR #523, opened 2025-05-16
[CI] Update CI workflow with new patch mechanism

phoenixdong:
1. Unpatch the corresponding package according to the type of test being run.
2. Adjust the import path of the package when running unit tests.
3. Skip the wrong test and mark it as TODO.


aoyulong (approved):
LGTM

PR #529, opened 2025-05-20
[Tools] Update patch method for hardware

Caozhou1995:
This PR updates the patching approach: instead of syncing back to the backends directory and applying patches there, it now directly diffs the submodule.


aoyulong (approved):
LGTM

PR #532, opened 2025-05-21
[CI][vllm] Reinstall vllm in workflow to synchronize code changes 

phoenixdong:
When running the test, compile and install vllm to synchronize code changes to ensure that the latest vllm code is tested.


aoyulong (approved):
LGTM

PR #534, opened 2025-05-21
[Tools] Support specified submodule commit for hardware patch

Caozhou1995:
This PR adds the ability to specify a commit for the submodule and optionally apply the FlagScale adaptation. Usage is as follows:
`python tools/patch/unpatch.py --backend vllm --backend-commit <vllm_commit> --no-fs-extension`
`--backend-commit` indicates the commit to which the submodule will automatically reset after a submodule update. The `--no-fs-extension` means the FlagScale adaptation will not be applied, and the submodule will remain in its original state.

### DIFF
--- a/tools/patch/git_utils.py
+++ b/tools/patch/git_utils.py
@@ -1,6 +1,5 @@
 from logger_utils import get_patch_logger
 
-
 logger = get_patch_logger()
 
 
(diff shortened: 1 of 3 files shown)


aoyulong (approved):
LGTM

PR #535, opened 2025-05-22
[Serve] Add SGLang as a new backend

Hchnr:
1. Support serve with SGLang
2. Support auto-tune with SGLang
3. Do some adaption on serve and auto-tune for SGLang, delete 6 common args
4. Add profile report for qwen3-0.6b with 3 backends


aoyulong (approved):
LGTM

PR #536, opened 2025-05-22
[Core] add use_gems(forward only) in vllm/v1/

shh2000:
This PR adds use_gems(forward only) in vllm/v1/


Caozhou1995:
Please use `# --- FLAGSCALE MODIFICATION BEG ---` and` # --- FLAGSCALE MODIFICATION END --- `to mark multi-line modifications, and use `# --- FLAGSCALE MODIFICATION ---` to indicate a single-line modification.

PR #539, opened 2025-05-22
[Serve] Support to clear task by action=stop cross multiple nodes

cyber-pioneer:
## Description
python run.py --config-path ./examples/deepseek_r1/conf --config-name serve **action=stop**


aoyulong (approved):
LGTM

PR #556, opened 2025-05-28
[BugFix] Fix commit topo order bug

Caozhou1995:
This PR fixed the bug of selecting the newest commit when backend commits is different.


ceci3 (approved):
LGTM

PR #644, opened 2025-07-04
Add the Support for DualPipeV

lxd-cumt:
**We support DualPipeV in FlagScale, an efficient pipeline scheduling method for MoE models.**
DualPipeV aims to address the long - standing issue of high All2All communication latency in MoE models.
Our experimental results show DualPipeV enhances training throughput by a factor of  4%-11%, across DeepSeek models of varying scales.

**Usage**
We provide two versions of DualPipeV: the base version and the optimized version. 
To run the base version, simply configure **use_dualpipev: true**. 
For the optimized comp/comm overlapping version, set **use_dualpipev: true** and **moe_fb_overlap: true**.

**Features**
Base Version:
- [x] Basic pipeline schedules, model of each pipeline stage is further split into two chunks, and the pipeline is arranged in a V-shape. Bubble rate is efficiently reduced based on this schedules.
- [x] Compatible with other distributed training parallelism strategies, including Tensor Parallelism (TP), Sequence Parallelism (SP), Data Parallelism (DP), and Context Parallelism (CP). 
截屏2025-07-03 15 08 22

Optimized Version:
- [x] Cross micro-batch computation-communication overlap during the steady phase.
- [x] Backward for weight / input is separated based on TransformerEngine, which can reduce the bubble ratio during the warmup and cooldown phases.

**Future Works**

- [x] Evaluate accuracy and performance on more MoE models 
- [x] Support hybrid parallel strategy, including tensor parallel, distributed data parallel, sequence parallel, context parallel, 
- [ ] Support multi-token prediction, support tie embeddings and lm head
- [ ] Support recompute
- [ ] More optimizations...

Somd modules of DualPipeV  references the code and design of [Ascend MindSpeed DualPipeV](https://gitee.com/ascend/MindSpeed/blob/master/mindspeed/core/pipeline_parallel/dualpipev/dualpipev_schedules.py)


flagscale/backends/Megatron-LM/megatron/core/parallel_state.py
```diff
@@ -2628,6 +2641,9 @@ def destroy_model_parallel():
     global _VIRTUAL_PIPELINE_MODEL_PARALLEL_WORLD_SIZE
     _VIRTUAL_PIPELINE_MODEL_PARALLEL_WORLD_SIZE = None
 
+    global _DUALPIPEV_PIPELINE_MODEL_PARALLEL_WORLD_SIZE
```
heavyrain-lzy, line 2644: The contents that FlagScale adds to the Megatron files, We can add a tag like 
```
######### FlagScale Begin ########
 xxxxxx
######### FlagScale End ########
```
or `######### FlagScale Modify ########` for only one line.

aoyulong (approved):
LGTM

PR #696, opened 2025-07-28
[CICD] Enhance Test Flow Error Detection and Update GPU Detection Method

phoenixdong:
1. Strict test flow failure detection, the goal is that any error, regardless of size, will throw an error/exception.
2. Update GPU detection method, the previous method could execute but would throw errors.

### DIFF
--- a/tests/scripts/_gpu_check.sh
+++ b/tests/scripts/_gpu_check.sh
@@ -1,40 +1,42 @@
 #!/bin/bash
-
 wait_for_gpu() {
+    local gpu_count
     gpu_count=$(nvidia-smi --query-gpu=name --format=csv,noheader | wc -l)
-
-    memory_usage_max=30000
-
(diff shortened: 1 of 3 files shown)


tests/scripts/_gpu_check.sh
```diff
@@ -1,40 +1,42 @@
...
-        fi
-
-    done
-
-    if [ "$need_wait" = false ]; then
+        # Use mapfile instead of read to avoid IFS issues and ensure proper array population
+        local memory_usage_array=()
+        local memory_total_array=()
+        # Query GPU memory usage and total memory, suppress stderr to prevent exit on failure
+        mapfile -t memory_usage_array < <(nvidia-smi --query-gpu=memory.used --format=csv,noheader,nounits 2>/dev/null)
```
aoyulong, line 11: What about other platforms besides NVIDIA?

PR #698, opened 2025-07-28
[CICD] Create Metax Dockerfile

zihugithub:
**New Dockerfile**

- **Dockerfile.metax:** Add Metax Image Automatic Build Process
- **install-metax.sh:** Add Metax image inference environment automatic build process


aoyulong (approved):
LGTM

PR #710, opened 2025-08-05
Standardize and polish DualPipeV related codes

lxd-cumt:
- move DualPipeV related codes to **flagscale/train/dualpipev/**
- add comments for Megatron-LM patches in FlagScale


heavyrain-lzy (approved):
LGTM

PR #712, opened 2025-08-05
[HeteroTrain] Fix some bugs in heterogeneous training.

heavyrain-lzy:
Fix some bugs in the latest hetero-train.
1. `get_expert_data_parallel_world_size` error problem because the corresponding hetero interface is not added

2. Solve the problem of hetero training on torch2.6

3. Solve the problem of dp heterogeneous communication error under shared_embedding because only the current group's rank calls this API at new_group


flagscale/train/hetero/parallel_context.py
```diff
@@ -757,7 +755,7 @@ def _backtrack(mesh_index, prev_rank, path, token = "pp", is_expert=False):
             if mesh_index == len(self._process_meshes):
                 aggregated_ranks = [rank for ranks in path for rank in ranks]
                 self._global_all_group_ranks[group_name].append(aggregated_ranks)
-                group = create_group(aggregated_ranks, timeout=self._timeout, use_local_synchronization=True, group_desc=group_name)
+                group = create_group(aggregated_ranks, timeout=self._timeout, use_local_synchronization=False, group_desc=group_name)
```
aoyulong, line 759: Please include some comments.

PR #742, opened 2025-08-18
[AutoTune] Fix some silent grep failures

shuailong616:
**Problem**
In the case of multiple nodes, the virtual environment of the slave node is not activated, and there is no correct error log.The automatic tuning process will not be interrupted, and no performance data is recorded.
没有性能数据
**Solution**
In the case of multiple nodes, prompt information will be given when the "cmds" attribute is absent, guiding the activation of the virtual environment on the slave node
错误提示


flagscale/runner/runner_train.py
```diff
@@ -193,6 +193,11 @@ def _generate_run_script_train(config, host, node_rank, cmd, background=True, wi
     if cmds_config:
         before_start = cmds_config.get("before_start", "")
     else:
+        nnodes = config.experiment.runner.get("nnodes", None)
+        if nnodes is not None and nnodes > 1:
+            raise ValueError(
+                f"Nndoes value is > 1, but 'cmd' configuration not found in experiment, please 'source /root/miniconda3/bin/activate flagscale-train' at slave node."
+            )
```
aoyulong, line 200: I believe it is the user's responsibility to ensure all nodes have a uniform environment.

Caozhou1995:
@gemini-code-assist please review this pull request in details.

Caozhou1995:
closed by it is the user's responsibility to ensure all nodes have a uniform environment.

PR #753, opened 2025-08-27
[Serve] Support qwen3 8b serving

ceci3:


cyber-pioneer:
Is the config file missing in examples/qwen3/conf?

aoyulong (approved):
LGTM

PR #772, opened 2025-09-08
[Megatron-LM] Fix save ckpt

heavyrain-lzy:
Fix the bug when saving checkpoints.
The tracker file is created by rank 0 but wandb_finalize_fn is called on the last rank.


aoyulong (approved):
LGTM

PR #776, opened 2025-09-10
[Feature] Add support for LoRA

lxd-cumt:
Add support for LoRA


flagscale/backends/Megatron-LM/megatron/training/arguments.py
```diff
@@ -1559,6 +1570,29 @@ def _add_network_size_args(parser):
                        'We compute the average of the MTP losses across all depths, '
                        'and multiply it the scaling factor to obtain the overall MTP loss, '
                        'which serves as an additional training objective.')
+    group.add_argument('--use-lora', action='store_true',
```
heavyrain-lzy, line 1573: It's better to use `######### FlagScale Begin/End ########` to enclose the code.

PR #807, opened 2025-09-18
[kunlunxin]adaption for qwen3-8b

brianlcy123:


Caozhou1995:
close by used latest patch

PR #818, opened 2025-09-22
[Autotuner] Fix the problem that can not find the performance data

shuailong616:
Sometimes, on the Kunlunxin chip's card, the performance result of the automatic tuning may be found on the last card of the first node instead of the last card of the final node.


flagscale/runner/auto_tuner/record/recorder.py
```diff
@@ -173,73 +173,72 @@ def grep_max_memory(self, path, pattern="max reserved"):
...
+                    metric = re.findall(metric_pattern, line)
+                    if metric:
+                        value = None
+                        for item in metric[0]:
+                            try:
+                                value = float(item)
+                                performance.append(value)
+                                break
+                            except:
+                                assert value is not None, "Can't grep the performance"
```
Caozhou1995: Fix the bug when except and value is none

shuailong616: fix

Caozhou1995 (approved):
LGTM

PR #827, opened 2025-09-23
[Runner] support served model name for autotune

shuailong616:
The current automatic tuning benchmark function only supports the mode parameter and does not support served_model_name, which leads to an error if served_model_name is used.

### DIFF
--- a/flagscale/runner/runner_serve.py
+++ b/flagscale/runner/runner_serve.py
@@ -1173,7 +1173,7 @@ class SSHServeRunner(RunnerBase):
 
         trust_remote_code = engine_args.get("trust_remote_code", False)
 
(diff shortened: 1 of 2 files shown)


flagscale/runner/runner_serve.py
```diff
@@ -1173,7 +1173,8 @@ def _profile_serve(self):
 
         trust_remote_code = engine_args.get("trust_remote_code", False)
 
-        model_name = engine_args.get("served_model_name", None) or engine_args.get("model", None)
+        model_id = engine_args.get("served_model_name", None)
```
cyber-pioneer: model_id 统一命名成served_model_name，避免语义混淆

cyber-pioneer:
LGTM

PR #831, opened 2025-09-23
Support Pi0 Inference

Hchnr:
Support Pi0 Inference 
- Add a new entrypoint for pi0 inference.
- Add an inference example.
- Add lerobot as a new backend.

### DIFF
--- a/tools/patch/patch.py
+++ b/tools/patch/patch.py
@@ -513,7 +513,7 @@ def normalize_backend(backend):
         return "omniinfer"
     elif input_lower in ["verl"]:
         return "verl"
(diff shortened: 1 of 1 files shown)


ceci3 (commented):
LGTM

ceci3 (dismissed):

tools/patch/patch.py
```diff
@@ -513,6 +513,8 @@ def normalize_backend(backend):
         return "omniinfer"
     elif input_lower in ["verl"]:
         return "verl"
+    elif input_lower in ["lerobot", "le-robot", "le_robot", "Lerobot"]:
```
cyber-pioneer: better to limit key?

cyber-pioneer:
LGTM

PR #845, opened 2025-09-28
[Autotune]  autotune space set error  generated some problems

shuailong616:
when  cards =2,if set tp =4 in space ,  generate strategies of tp = 32
output


Caozhou1995 (approved):
LGTM

PR #876, opened 2025-10-23
[CICD] Add sccache caching for inference test cases to accelerate VLLM compilation

zihugithub:
Add VLLM compilation cache SCCache for unit test inference cases


.github/workflows/unit-tests-nvidia.yml
```diff
@@ -66,9 +66,85 @@ jobs:
...
+            export SCCACHE_DIR=/root/.cache/sccache
+            sccache --start-server
+            sccache --show-stats
+
+            echo "[INFO] Install VLLM and its dependencies ..."
+            MAX_RETRIES=5
+            DELAYS=(1 1.5 2 2.5 3)
+            COMMANDS=(
+              "pip install scikit-build scikit-build-core"
+              "pip install git+https://github.com/FlagOpen/FlagGems.git@v3.0"
```
legitnull, line 77: Do we need a separate file to explicitly list the version of the dependencies?

Caozhou1995, line 77: Why not install scikit-build and scikit-build-core directly into the image

zihugithub: Installing scikit build and scikit build core in the image has been optimized. Thank you for your feedback

zihugithub: Received, thank you for your feedback

PR #882, opened 2025-10-29
[Elastic] Fix log collection and diagnostic for multi-node

wanglei19991004:
Fixed the issue where switches do not take effect under multi-machine operation.

### DIFF
--- a/examples/qwen2_5/conf/train/10b.yaml
+++ b/examples/qwen2_5/conf/train/10b.yaml
@@ -1,0 +1,88 @@
+system:
+  tensor_model_parallel_size: 1
+  pipeline_model_parallel_size: 2
+  context_parallel_size: 1
+  disable_bias_linear: true
+  reset_position_ids: True
+  reset_attention_mask: True
+  qk_layernorm: false
+  sequence_parallel: true
+  use_distributed_optimizer: true
+  overlap_grad_reduce: true
+  overlap_param_gather: true
+  finetune: true
+  precision:
+    bf16: true
+    attention_softmax_in_fp32: true
+    accumulate_allreduce_grads_in_fp32: true
+  logging:
+    log_interval: 1
+    tensorboard_log_interval: 1
+    wandb_project: ${experiment.exp_name}
+    wandb_exp_name: ${experiment.exp_name}
+    log_timers_to_tensorboard: true
+    log_validation_ppl_to_tensorboard: true
+    log_throughput: true
+    log_params_norm: true
+    log_num_zeros_in_grad: true
+    log_memory_to_tensorboard: true
+  checkpoint:
+    save_interval: ${experiment.save_steps}
+    load: ${experiment.load}
+    ckpt_format: ${experiment.ckpt_format}
+
+model:
+  transformer_impl: transformer_engine
+  num_layers: 52
+  hidden_size: 4096
+  ffn_hidden_size: 11008
+  kv_channels: 128
+  group_query_attention: true
+  num_attention_heads: 32
+  num_query_groups: 4 # num_key_value_heads
+  seq_length: 4096
+  max_position_embeddings: 32768
+  norm_epsilon: 1e-6
+  use_rotary_position_embeddings: true
+  rotary_base: 1000000
+  swiglu: true
+  normalization: RMSNorm
+  init_method_std: 0.02
+  attention_dropout: 0.0
+  hidden_dropout: 0.0
+  clip_grad: 1.0
+  position_embedding_type: rope
+  untie_embeddings_and_output_weights: true
+  no_position_embedding: true
+  no_rope_fusion: true
+
+  # training
+  seed: 42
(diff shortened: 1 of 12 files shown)


flagscale/runner/elastic/monitor_launcher.py
```diff
@@ -54,6 +54,8 @@ def main():
     parser.add_argument(
         "--pid-file", required=True, help="The path of the PID file for the training process"
     )
+    parser.add_argument("--host", required=True, help="Hostname or IP of this node")
+    parser.add_argument("--node-rank", type=int, required=True, help="Node rank of this node")
```
tengqm, line 56: These two arguments are both required ... that means the launcher can be used only for single-node scenario. Am I understanding this correctly?

flagscale/runner/runner_train.py
```diff
@@ -487,8 +489,11 @@ def run(
         interval=10,
         enable_log_collection=True,
         enable_diagnostic=True,
-        enable_monitoring=False,
+        enable_monitoring=None,
     ):
+        # Read from config if not explicitly provided
+        if enable_monitoring is None:
+            enable_monitoring = self.config.experiment.runner.get("enable_monitoring", True)
```
tengqm: This is a behavior change ...
The monitoring was defaulted to False previously, now it is defaulted to True.
Is this alright?

flagscale/runner/elastic/monitor_launcher.py
wanglei19991004: These two arguments are used to generate a separate diagnostic file for each host + node pair.
The current implementation actually supports multi-host and multi-node scenarios, not just a single node.

flagscale/runner/runner_train.py
wanglei19991004: Sorry！It should be defaulted to False, and I will fix it accordingly.

flagscale/runner/utils.py
```diff
@@ -213,6 +213,66 @@ def run_ssh_command(host, cmd, port=None, dryrun=False, query=False):
...
+    """
+    try:
+        cmd = ["ssh", "-p", str(port), host, f"stat -c%Y {filepath}"]
+        result = subprocess.run(cmd, capture_output=True, text=True, check=True)
+        return int(result.stdout.strip())
+    except subprocess.CalledProcessError:
+        return -1
+
+
+def get_remote_file_exists(host, filepath, port=22):
```
tengqm: This function is added but not used, right?

wanglei19991004: Yeah~ I will remove it accordingly.

zhaoyinglia (approved):
LGTM

flagscale/runner/elastic/log_collector.py
```diff
@@ -115,34 +86,32 @@ def collect_logs(config, host, node_rank, destination_dir, dryrun=False):
     no_shared_fs = config.experiment.runner.get("no_shared_fs", False)
     log_dir = logging_config.log_dir
     src_log_file = find_actual_log_file(log_dir, node_rank, host, no_shared_fs)
-    dest_log_file = os.path.join(
-        destination_dir,
-        f"host_{node_rank}_{host}_temp_{datetime.now().strftime('%Y%m%d_%H%M%S')}.log",
-    )
+    dest_log_file = os.path.join(destination_dir, f"host_{node_rank}_{host}_current.log")
```
tengqm, line 89: shall we remove `_current` from the name?

flagscale/runner/elastic/monitor_launcher.py
tengqm: My question was that whether the two arguments are ALWAYS REQUIRED.

flagscale/runner/elastic/monitor_service.py
```diff
@@ -1,30 +1,31 @@
...
+    def __init__(self, config, runner_instance, interval=10, host=None, node_rank=None):
         """
         Initializing service
 
         Args:
             config: Configuration object
             runner_instance: runner instance
             interval: interval time
+            host: Hostname or IP of this node (for single-node mode)
+            node_rank: Node rank of this node (for single-node mode)
```
tengqm, line 28: These two arguments clearly indicate that neither `host` nor `node_rank` is required for single-node mode.

flagscale/runner/elastic/log_collector.py
zhaoyinglia: FYI @wanglei19991004

flagscale/runner/elastic/monitor_service.py
wanglei19991004: The "single-node mode" means that in a multi-machine environment with no_shared_fs, each machine starts a process separately for monitoring. However, in the development and testing environment, only a shared file system (shared_fs) is available, and the functionality has not been tested for correctness. Therefore, there might be issues with no_shared_fs.

PR #885, opened 2025-10-29
Add Seq Pack to verl in third_party

WenLiuyi:


examples/qwen2_5/conf/rl/0_5b.yaml
```diff
@@ -20,8 +20,8 @@ data:
 actor_rollout_ref:
   model:
     path: /workspace/models/Qwen/Qwen2.5-0.5B-Instruct
-    use_remove_padding: true
```
heavyrain-lzy: Why delete this?

flagscale/runner/utils.py
```diff
@@ -256,10 +256,8 @@ def flatten_dict_to_args_verl(config_dict, pre_str=""):
                 target_str = f"{key}."
             args.extend(flatten_dict_to_args_verl(value, pre_str + target_str))
         elif isinstance(value, list):
-            v_str = ""
-            for v in value:
-                v_str += f"{v}"
```
heavyrain-lzy: Why change this?

flagscale/backends/verl/verl/utils/seqlen_balancing.py.patch
```diff
@@ -0,0 +1,827 @@
...
+-    partitions = final_state.get_partitions()
+-    if equal_size:
+-        for i, partition in enumerate(partitions):
+-            assert len(partition) * k_partitions == len(seqlen_list), (
+-                f"{len(partition)} * {k_partitions} != {len(seqlen_list)}"
+-            )
+-    return partitions
+-
+-
+-def greedy_partition(seqlen_list: list[int], k_partitions: int, equal_size: bool):
```
heavyrain-lzy, line 328: It is best to keep the original functions and only add new functions without deleting the original functions.

examples/qwen2_5/conf/rl/0_5b.yaml
WenLiuyi: use_remove_padding is only relevant to FSDP, which controls whether to enable seqpack optimization.  For Megatron, remove padding is a built-in design, not controlled by the parameter 'use_remove_padding', which leads to 'use_remove_padding' to be redundant.

flagscale/runner/utils.py
WenLiuyi: In experiments, if there were multiple elements in logger, supposing: logger: ["console", "wandb", "tensorboard"], it would be converted to concatenated string: ""consolewandbtensorboard", not a list 

An undiscovered problem

flagscale/backends/verl/verl/utils/seqlen_balancing.py.patch
WenLiuyi: ok., it's not deleted. original function greedy_partition and karmarkar_karp has been integrated into PackingAlgorithm for unified management.

PR #899, opened 2025-11-03
[Tools] Refine git apply logic in unpatch.py to ensure patch hash consistency

yanqingliu35-stack:
unpatch.py currently always uses git apply --whitespace=fix, which alters patch hashes by "fixing" whitespace  This causes an index mismatch when running patch after unpatch. This PR fixes it by first attempting a strict git apply, and only falling back to git apply --whitespace=fix if the strict apply fails. This preserves the correct patch hash while maintaining compatibility.

### DIFF
--- a/tools/patch/unpatch.py
+++ b/tools/patch/unpatch.py
@@ -276,13 +276,13 @@ def apply_hardware_patch(
 
             repo = Repo(submodule_path)
             try:
-                repo.git.apply(new_patch_file)
+                repo.git.apply("--whitespace", "warn", new_patch_file)
             except Exception as e:
                 logger.warning(
                     f"Failed to apply patch cleanly, and error is {e.stderr}. Retrying with --whitespace=fix."
                 )
 
-                repo.git.apply("--whitespace", "warn", new_patch_file)
+                repo.git.apply("--whitespace", "fix", new_patch_file)
             logger.info(f"Patch {new_patch_file} has been applied.")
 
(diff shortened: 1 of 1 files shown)


tools/patch/unpatch.py
```diff
@@ -275,7 +275,13 @@ def apply_hardware_patch(
             )
 
             repo = Repo(submodule_path)
-            repo.git.apply("--whitespace", "fix", new_patch_file)
+            try:
+                repo.git.apply(new_patch_file)
+            except git.exc.GitCommandError as e:
+                logger.warning(
+                    f"    Failed to apply patch cleanly. Retrying with --whitespace=fix. Error: {e.stderr}"
```
Caozhou1995: Error first

tools/patch/unpatch.py
```diff
@@ -275,8 +275,15 @@ def apply_hardware_patch(
             )
 
             repo = Repo(submodule_path)
-            repo.git.apply("--whitespace", "fix", new_patch_file)
-            logger.info(f"    Patch {new_patch_file} has been applied.")
+            try:
+                repo.git.apply(new_patch_file)
```
heavyrain-lzy: Using `repo.git.apply("--whitespace", "warn", new_patch_file)` to avoid default configuraion error of git.

tools/patch/unpatch.py
yanqingliu35-stack: Done

tools/patch/unpatch.py
yanqingliu35-stack: Done

PR #919, opened 2025-11-12
[Huawei-npu] Support qwen2.5-Vl for huawei NPU

heavyrain-lzy:
### PR Category

Train
### PR Types

Others
### PR Description

Support to train qwen2.5-Vl using Huawei NPU.


hardware/Huawei_Atlas800TA3/FlagScale/run.sh.patch
```diff
@@ -1,6 +1,6 @@
 diff --git a/run.sh b/run.sh
 new file mode 100644
-index 000000000..072a78c94
+index 00000000..072a78c9
 --- /dev/null
 +++ b/run.sh
```
Caozhou1995: Do we really need this? If necessary, it will be better to move it to examples/qwen2_5_vl/conf

heavyrain-lzy: > Do we really need this? If necessary, it will be better to move it to examples/qwen2_5_vl/conf

Just a sample. I'll delete it afterwards.

Caozhou1995 (dismissed):
LGTM

PR #923, opened 2025-11-17
[Serve] request args support general standard types

cyber-pioneer:
### PR Category

Serve

### PR Types

Improvements

### PR Description

request args support general standard types like builtin and typing types like:
str
int
bool
float
Optional[str]
Optional[int]
List[str]
List[int]
List[Dict[str, str]]
Dict[str, str]
Dict[str, List[int]]
Tuple[int, str]

ci result:
```shell
tests/scripts/functional_tests/test_task.sh --type serve --task base
``` 
image

### DIFF
--- a/flagscale/serve/engine.py
+++ b/flagscale/serve/engine.py
@@ -9,7 +9,6 @@ import os
 import subprocess
 import sys
 import threading
-import typing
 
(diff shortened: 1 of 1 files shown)


ceci3 (approved):
LGTM

PR #930, opened 2025-11-22
[Tools] Patch Mechanism Upgrade

yanqingliu35-stack:
This pr upgrades the patch mechanism, separating the modifications made by hardware manufacturers from the basic modifications of flagscale

### DIFF
--- a/tools/patch/patchv3.py
+++ b/tools/patch/patchv3.py
@@ -1,0 +1,589 @@
+# -*- coding: utf-8 -*-
+import argparse
+import copy
+import os
+import re
+import shutil
+import tempfile
+
+import git
+import yaml
+
+from encryption_utils import encrypt_file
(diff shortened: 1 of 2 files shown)


tools/patch/patchv3.py
```diff
@@ -0,0 +1,589 @@
...
+    get_file_statuses_for_untracked,
+    get_submodule_commit,
+)
+from logger_utils import get_patch_logger
+
+FLAGSCALE_BACKEND = "FlagScale"
+logger = get_patch_logger()
+
+
+def generate_and_save_patch(sub_repo, base_commit, file_path, status, src_dir):
```
tengqm: Please add some description about the parameters in the docstring.
For example, another user may get confused what is a `sub_repo`.

PR #933, opened 2025-11-25
Add the Megatron-Bridge

chai-xiaonan:
To address the current limitation where Megatron does not support saving and loading model weights in the Hugging Face (HF) safetensor format, the Megatron-Bridge functionality has been integrated. This enables the framework to support warm starts and saving using the HF safetensor model weight format. However, it does not support saving or loading the optimizer state dictionary. When using the HF safetensor format for warm starts, training always begins from iteration 0 due to the absence of optimizer states.
Additionally, two arguments have been added to enable the use of the HF format, one of which initializes the model from the HF config format. The tokenizer has been modified to resolve shape inconsistencies between HF weight loading and the MCore model. Furthermore, new options have been introduced in the transformer configuration to allow Megatron-Bridge to access the model's parameters.

### DIFF
--- a/examples/qwen2_5/conf/train/10b.yaml
+++ b/examples/qwen2_5/conf/train/10b.yaml
@@ -1,0 +1,88 @@
+system:
+  tensor_model_parallel_size: 1
+  pipeline_model_parallel_size: 2
+  context_parallel_size: 1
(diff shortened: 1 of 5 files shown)


chai-xiaonan:
收到，谢谢

PR #986, opened 2025-12-22
Rework the README for the project

tengqm:
### PR Category

Documentation

### PR Types

Docs

### PR Description

This PR reworks the README file for clarity improvement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,292 +1,123 @@
 
 [<img width="4750" height="958" alt="github+banner__2025-11-11+13_27_10" src="https://github.com/user-attachments/assets/e63014d8-ac72-4b82-98f7-aeed9833672a" />](https://www.flagopen.ac.cn/)
 
+<!--TODO(Qiming): Leave a placeholder for announcements.-->
 
-## 🔥 Latest News
+## About
 
-- **[2025/09]** Released [v0.9.0](https://github.com/FlagOpen/FlagScale/tree/release/v0.9.0):
-  - Training & Finetuning: Added LoRA for efficient finetuning, improved the autotuner for cross-chip heterogeneous training, and enabled distributed RWKV training.
-  - Inference & Serving: Introduced DiffusionEngine for FLUX.1-dev, Qwen-Image, and Wan2.1-T2V, support multi-model automatic orchestration and dynamic scaling.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -1,292 +1,122 @@
 
 [<img width="4750" height="958" alt="github+banner__2025-11-11+13_27_10" src="https://github.com/user-attachments/assets/e63014d8-ac72-4b82-98f7-aeed9833672a" />](https://www.flagopen.ac.cn/)
 
-
+<!--TODO(Qiming): Leave a placeholder for announcements.
 ## 🔥 Latest News
+[Changelog](./docs/CHANGELOG.md)
```
Darryl233: seems Changelog failed to be displayed in the README? And confirm whether the Changelog is necessary. Refer to frameworks like vllm, which also don't use a Changelog.

tengqm: Thanks for the review.

In an ideal world, the changelog should be auto-generated and auto-maintained using the change management toolchain. These information are valuable for users to assess whether an upgrade is necessary and how the overall project is evolving.

We are still striving to get those facilities in place. So ... the first step is to cleanse the README file, and the next step is to figure out how the `/docs` directory layout can be improved, and the next step is to see if some docs can be consolidated into a central repository ... This is simply a babystep towards that end.

tengqm: Also ... line 4-7 in the README is currently commented out.
I was leaving this TODO as a future work. The link is just a reminder for me that there was something to consider. :)

Darryl233: I totally agree that auto-generating changelogs via a toolchain benefits users in assessing upgrades and tracking project evolution. However, perhaps before we actually put the change management toolchain into use, we should keep the original Latest News section? This way, users can still access the latest update information in the first place.

tengqm: My bad. I forgot to mention that I'm proposing some follow-up PRs to make sure users can still check the change log. We will make sure the log is accessible until we find a better alternative to auto-generate it.

tengqm: Okay, now I've restored the link to the CHANGELOG in a resource section.
Some more revisions are still ongoing, this revised version is not the final version yet but I believe it is a self-contained babystep towards the goal.

PR #1007, opened 2025-12-31
Added the megatron-bridge tool to support loading and saving HF files.

chai-xiaonan:
The Nemo-Bridge model has been adapted to FlagScale, enabling FlagScale to support saving and loading checkpoints in HF safe tensor format. Verification was performed on Qwen3-0.6B, Deepseek v3-16_a3B, and Qwen-32B models; saving and loading HF safe tensor format worked without issues, and the accuracy was correct.


lxd-cumt:
Please add an argument, hf-save-steps, to dynamically control how often to save an Hugging Face checkpoint during training.

lxd-cumt:
Please remove patches, and pr to Megatron-LM-FL for megatron/core related modification

PR #1039, opened 2026-01-06
Adaptation of Qwen3-VL and Qwen2.5-VL

Hchnr:
### PR Category
Train

### PR Types
Bug Fixes

### PR Description
- Adaptation of Qwen3-VL for new Megatron.
- Adaptation of Qwen2.5-VL for new Transformers.


flagscale/models/megatron/qwen3_vl/vision_attention.py
```diff
@@ -318,6 +322,7 @@ def forward(
         sequence_len_offset: Optional[int] = None,
         *,
         inference_params: Optional[BaseInferenceContext] = None,
+        rotary_pos_cos_sin=None,
```
heavyrain-lzy, line 325: What will happen if we don't add this argument? Will it  cause an error?

heavyrain-lzy: > What will happen if we don't add this argument? Will it cause an error?

Sorry, I get it.

PR #1047, opened 2026-01-08
[DOC] Add a document for RL

heavyrain-lzy:
### PR Category

RL
### PR Types

Docs
### PR Description

Add the document of RL to run the example.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -68,7 +68,7 @@ We recommend using the latest release of [NGC's PyTorch container](https://catal
     cd verl-FL
     pip install --no-deps -e .
     ```
-    See more details in [verl-FL](https://github.com/flagos-ai/verl-FL.git)
+    See more details in [verl-FL](https://github.com/flagos-ai/verl-FL.git) to get full installation instructions.
 
 
 ## Run a Task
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -256,3 +267,64 @@ flagscale serve <MODEL_NAME> <MODEL_CONFIG_YAML>
 
 The configuration files allow you to specify the necessary parameters and settings
 for your deployment, ensuring a smooth and efficient serving process.
+
+### RL
```
Darryl233, line 225: I suggest putting RL section after Serve, making all the examples of qwen3 be together
and keep Serve Deepseek-R1 at the end

docs/getting-started.md
```diff
@@ -60,6 +60,17 @@ We recommend using the latest release of [NGC's PyTorch container](https://catal
     ```
     See more details in [Megatron-LM-FL](https://github.com/flagos-ai/Megatron-LM-FL)
 
+
+- **RL backend**
+    verl-FL: 
+    ```sh
+    git clone https://github.com/flagos-ai/verl-FL.git
+    cd verl-FL
+    pip install --no-deps -e .
```
Darryl233, line 69: With “--no-deps” option，pip will not automatically install verl-FL dependencies. Is that ok?

heavyrain-lzy: It is recommended to install dependencies separately in Verl.

docs/getting-started.md
heavyrain-lzy: Thanks. I will change it.

docs/getting-started.md
Darryl233: got it

PR #1055, opened 2026-01-11
[CodeStyle] Update FS codestyle from black to ruff

zhaoyinglia:
### PR Category
Others

### PR Types
Others

### PR Description
- [x] update codestyle from black to ruff
- [x] add `typo` check
- [ ] add max length check
- [ ] add variable type check with `mypy`
- [ ] add copyright

Results in local env:
image

Results in CI env:
image

### DIFF
--- a/flagscale/runner/backend/backend_vllm.py
+++ b/flagscale/runner/backend/backend_vllm.py
@@ -1,5 +1,7 @@
+import collections
 import json
 import os
+import socket
 
 from omegaconf import DictConfig, OmegaConf
 
@@ -7,9 +9,14 @@ from flagscale.runner.backend.backend_base import BackendBase
 from flagscale.runner.utils import (
     ResourceManager,
     flatten_dict_to_args,
+    get_addr,
     get_free_port,
+    get_ip_addr,
+    is_ip_addr,
+    is_master_node,
     logger,
     parse_hostfile,
+    wait_for_ray_master,
 )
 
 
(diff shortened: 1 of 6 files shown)


.pre-commit-config.yaml
```diff
@@ -1,74 +1,46 @@
...
-        .*\.md|
-        .*\.sh|
-        .*\.txt|
-        .*\.yaml|
-        .*\.yml$
+        flagscale/train/train.*\.py|
+        flagscale/models/megatron/.*|
+        tools/checkpoint/loader.*\.py|
+        tools/checkpoint/saver.*\.py|
+        examples/aquila/tokenizer*|
```
tengqm, line 8: Why we exclude these specific source files?
Some comments would be very helpful for future maintainers.

pyproject.toml
```diff
@@ -1,11 +1,18 @@
 [build-system]
-requires = ["setuptools>=77.0", "wheel", "gitpython", "pyyaml", "cryptography", "pip", "hatchling", "hatch-vcs", "editables"]
+requires = [
+    "setuptools>=77.0",
+    "wheel",
+    "pip",
```
tengqm: Can/shall we pin the versions for wheel and pip?
Pinning version would help ensure the build successes/failures can be reproduced.

pyproject.toml
```diff
@@ -14,19 +21,105 @@ Homepage = "https://github.com/flagos-ai/FlagScale"
...
+    "RUF024",
+    "RUF026",
+    "RUF100",
+    # flake8-quotes
+    "Q003",
+    "Q004",
+]
+ignore = [
+    "E402",         # Module level import not at top of file
+    "E501",         # Line too long
```
tengqm, line 88: In the future, we may want to impose some constraints on long lines (though not necessarily 80 chars per line).

.pre-commit-config.yaml
zhaoyinglia: These files need to be regularly aligned with megatron-lm, so it is best to keep their format as similar as possible to the original.

pyproject.toml
zhaoyinglia: OK

pyproject.toml
zhaoyinglia: DONE.

PR #1073, opened 2026-01-18
Improve the workflows and refactor the install process

aoyulong:
### PR Category

CICD

### PR Types

Improvemetns

### PR Description

This pull request significantly refactors the GitHub Actions workflows for functional and unit tests, introducing improved modularity, flexibility, and robustness in the CI/CD pipeline. The main changes include splitting functional tests by task type, enhancing configuration loading, supporting Conda environments, and making artifact download more reliable.

**Workflow modularization and configuration improvements:**

* The `hardware` input is renamed to `platform` throughout the workflows for clarity and consistency. Configuration loading is refactored to use a dedicated script (`load_platform_config.sh`), and outputs are expanded to include separate test matrices for different tasks (training, heterogeneous training, inference, RL). 

* The main workflow now splits functional tests into multiple jobs by task type: `functional_tests_train`, `functional_tests_hetero_train`, `functional_tests_inference`, and `functional_tests_rl`. Each job uses its own workflow file and test matrix, allowing for more granular control and reporting. The test completion check is updated to handle the new job structure.

**Functional test workflow enhancements:**

* The `functional_tests_common.yml` workflow is updated to serve as a template, with new optional inputs for `conda_env` and `conda_path` to support running tests in different Conda environments. The workflow now sets up the environment more flexibly, using a utility script for Conda activation. 

* Artifact download steps are made more robust by retrying up to three times and verifying success, reducing flakiness due to transient network issues.

* The install and test execution steps are refactored to use environment variables, support Conda activation if requested, and display Python environment info for debugging. The test runner command is simplified and made more flexible.

These changes collectively make the CI/CD pipeline more maintainable, extensible, and reliable, especially as more test types and platforms are added.

### DIFF
--- a/requirements/cuda/base.txt
+++ b/requirements/cuda/base.txt
@@ -2,4 +2,3 @@
 # Includes PyTorch with CUDA support and common dependencies
 
 -r ../common.txt
-torch
--- a/tools/install/install.sh
+++ b/tools/install/install.sh
@@ -24,6 +24,7 @@ PLATFORM="cuda"  # Default to CUDA platform
 RETRY_COUNT="3"
 CONDA_ENV=""      # Optional: conda environment to activate
 CONDA_PATH=""     # Optional: custom conda installation path
+DEV_MODE="false"  # Install development dependencies (build, lint, test)
 
 # Dynamically discover valid tasks from platform configuration
 discover_valid_tasks() {
@@ -123,6 +124,7 @@ OPTIONS:
     --retry-count N          Number of retry attempts (default: 3)
     --conda-env ENV          Optional: activate conda environment before install
     --conda-path PATH        Optional: custom conda installation path
+    --dev                    Install development dependencies (build, lint, test)
     --help                   Show this help message
 
 EXAMPLES:
@@ -134,6 +136,9 @@ EXAMPLES:
 
     # Install all task dependencies
     $0 --task all --platform cuda
+
+    # Install with development dependencies (includes build, lint, test)
+    $0 --task train --dev
 
     # Install with custom retry count
     $0 --task train --retry-count 5
@@ -174,6 +179,10 @@ parse_args() {
             --conda-path)
                 CONDA_PATH="$2"
                 shift 2
+                ;;
+            --dev)
+                DEV_MODE="true"
+                shift
                 ;;
             --help|-h)
                 usage
@@ -258,7 +267,18 @@ install_base_dependencies() {
 # Install task-specific pip requirements
 install_task_requirements() {
     local task=$1
-    local requirements_file="$PROJECT_ROOT/requirements/$PLATFORM/${task}.txt"
+    local requirements_file
+
+    # Use _dev.txt if --dev flag is set, otherwise use regular .txt
+    if [ "$DEV_MODE" = "true" ]; then
+        requirements_file="$PROJECT_ROOT/requirements/$PLATFORM/${task}_dev.txt"
+        if [ ! -f "$requirements_file" ]; then
+            log_warn "Dev requirements not found: $requirements_file, falling back to regular"
+            requirements_file="$PROJECT_ROOT/requirements/$PLATFORM/${task}.txt"
+        fi
+    else
+        requirements_file="$PROJECT_ROOT/requirements/$PLATFORM/${task}.txt"
+    fi
 
     if [ ! -f "$requirements_file" ]; then
         log_info "No task requirements file: $requirements_file (skipping)"
@@ -266,6 +286,9 @@ install_task_requirements() {
     fi
 
     log_step "Installing pip requirements for task: $task"
+    if [ "$DEV_MODE" = "true" ]; then
+        log_info "Installing development dependencies (includes build, lint, test)"
+    fi
     retry_pip_install "$requirements_file" "$RETRY_COUNT"
 }
 
@@ -324,6 +347,11 @@ main() {
     # Display current environment
     log_info "Current conda environment: $(get_conda_env)"
     check_python_version || log_warn "Python version check failed (continuing anyway)"
+
+    # Display dev mode status
+    if [ "$DEV_MODE" = "true" ]; then
+        log_info "Development mode: ENABLED (will install build, lint, test deps)"
+    fi
 
     # Install dependencies based on task
     if [ "$TASK" = "all" ]; then


requirements/cuda/base.txt
```diff
@@ -0,0 +1,5 @@
+# CUDA platform-specific base dependencies
+# Includes PyTorch with CUDA support and common dependencies
```
tengqm, line 2: I'd suggest we maintain ALL the dependencies using the `pyproject.toml`, which serves as a single source of truth. Spreading dependencies into 10+ files may lead to sub-optimal maintainability when compared to the centralized approach.
The selective installation of these dependencies can be done using `pip` while users/developers always have a global view of the package list.

aoyulong: @tengqm That's a valid point. FlagScale should function as a meta framework, allowing users to install only the parts they need. Additionally, the design is inspired by <a href="https://github.com/vllm-project/vllm/tree/main/requirements">vllm</a>.

Darryl233: As a meta-framework, I think that the refactored FlagScale is a more lightweight repository compared to before, and it may not involve such complex dependency management at present. I quite agree with your idea, but perhaps we should further design and optimize the dependency management according to the needs of subsequent iterations, retaining only the necessary parts for now, rather than leaving too many redundant designs from the start.

.github/workflows/functional_tests_common.yml
```diff
@@ -51,12 +59,46 @@ jobs:
...
+      - name: Download source code artifact (attempt 3)
+        if: steps.download_attempt_2.outcome == 'failure'
         uses: actions/download-artifact@v4
+        id: download_attempt_3
         with:
           name: ${{ inputs.source_artifact }}
           path: /tmp
 
+      - name: Verify artifact download
+        shell: bash
```
zihugithub: Suggest moving shell: bash from each run in steps to the jobs.defaults.run.shell: bash level for better consistency and cleaner configuration.

.github/workflows/functional_tests_common.yml
```diff
@@ -80,54 +148,34 @@ jobs:
           TASK='${{ inputs.task }}'
           MODEL='${{ inputs.model }}'
           CASE='${{ inputs.case }}'
+          CONDA_ENV='${{ inputs.conda_env }}'
```
zihugithub, line 158: Defining these variables using the env block is more visually appealing and maintainable.

.github/workflows/functional_tests_common.yml
zihugithub: Use working-directory: /__w/FlagScale/FlagScale to explicitly set the working directory to the project root.

.github/workflows/functional_tests_common.yml
```diff
@@ -80,54 +148,34 @@ jobs:
...
           echo "Task: $TASK"
           echo "Model: $MODEL"
           echo "Case: ${CASE:-all}"
           echo "Project root: $PROJECT_ROOT"
 
-          # Load retry functions
-          source "$PROJECT_ROOT/.github/workflows/scripts/retry_functions.sh"
-
-          # Setup backend
-          source /root/miniconda3/bin/activate flagscale-train
```
zihugithub, line 97: Consider the scenario where the Conda installation location might change.

.github/workflows/functional_tests_common.yml
zihugithub: runs-on: ${{ fromJson(inputs.runs_on) }}
    defaults:
      run:
        shell: bash
    container: ....

xmhubj (dismissed):

requirements/cuda/base.txt
aoyulong: @Darryl233 I agree that we should keep the FlagScale dependency simple. I will add all.txt to consolidate the dependency in one place. However, the decomposition should align with workflows, installation, and requirements.

.github/workflows/functional_tests_common.yml
aoyulong: They are not environment variables but arguments provided by users.

.github/workflows/all_tests_common.yml
```diff
@@ -138,48 +99,135 @@ jobs:
...
-      platform: ${{ inputs.hardware }}
+      platform: ${{ inputs.platform }}
       device: ${{ matrix.device }}
       image: ${{ needs.checkout_and_config.outputs.ci_image }}
       runs_on: ${{ needs.checkout_and_config.outputs.runs_on }}
       container_volumes: ${{ needs.checkout_and_config.outputs.container_volumes }}
       container_options: ${{ needs.checkout_and_config.outputs.container_options }}
       source_artifact: flagscale-source-${{ github.sha }}
+      conda_env: flagscale-train  # Optional: can be empty for non-conda environments
+      conda_path: "/root/miniconda3"  # Optional: specify custom conda path, empty for auto-detection
```
tengqm, line 110: Actually, I'm not a big fan of conda. Conda is kind of a virtual environment just like venv or a docker container. I don't think we need three different kinds of isolation mechanisms in this project. If we are proceeding with the container approach, we can drop conda completely. This is from a perspective of "why are we using conda". Another factor to consider is that conda is heavy-weight. My previous analysis told me that the footprint of a conda deployment amounts to 680MB, which is non-trivial.

aoyulong: @tengqm For deployment, I agree with your point. However, for development, I think we can keep conda-like virtual environments optional since AI developers are familiar with using them.

tengqm: Okay. For AI developers, they are more familiar with conda environment than docker containers, right? We then get a serious question ... does it really make a lot of senses to push container-based development? Do we want/need to put development toolchain into containers?

aoyulong: @tengqm I believe there is no need to determine which option is better, as both serve their purposes. In real development, multiple users may share a single machine (containerized), and using different conda-like environments can help isolate their various changes. For deployment, we can eliminate the conda requirements since everything has been tested thoroughly.

tengqm: That is exactly what I'm trying to understand/clarify. Docker containers are for deployment scenarios, while conda environments are mostly for development scenarios. If this is true, we don't have a strong need for `-devel` flavor of container images. Correct?

.github/workflows/functional_tests_common.yml
```diff
@@ -51,83 +70,135 @@ jobs:
...
+            exit 1
+          fi
+
       - name: Extract source code
-        shell: bash
         run: |
-          mkdir -p /__w/FlagScale/FlagScale
-          tar -xzf /tmp/flagscale-source.tar.gz -C /__w/FlagScale/FlagScale
+          mkdir -p $PROJECT_ROOT
+          tar -xzf /tmp/flagscale-source.tar.gz -C $PROJECT_ROOT
```
tengqm, line 115: `flagscale-source.tar.gz` here should be `${{ inputs.source_artifact }}`

.github/workflows/functional_tests_common.yml
```diff
@@ -51,83 +70,135 @@ jobs:
...
           CASE='${{ inputs.case }}'
-          PROJECT_ROOT="/__w/FlagScale/FlagScale"
+          CONDA_ENV='${{ inputs.conda_env }}'
+          CONDA_PATH='${{ inputs.conda_path }}'
 
-          echo "Running functional tests"
+          echo "Running functional tests for $TASK"
           echo "Platform: $PLATFORM"
           echo "Device: $DEVICE"
           echo "Task: $TASK"
```
tengqm, line 164: This line can be removed, since this info already output from line 161.

.github/workflows/functional_tests_common.yml
```diff
@@ -23,6 +27,16 @@ on:
         required: false
```
tengqm, line 27: for an optional parameter, better provide a default value.

.github/workflows/functional_tests_common.yml
```diff
@@ -51,83 +70,135 @@ jobs:
...
+          CONDA_ENV='${{ inputs.conda_env }}'
+          CONDA_PATH='${{ inputs.conda_path }}'
 
-          echo "Running functional tests"
+          echo "Running functional tests for $TASK"
           echo "Platform: $PLATFORM"
           echo "Device: $DEVICE"
           echo "Task: $TASK"
           echo "Model: $MODEL"
           echo "Case: ${CASE:-all}"
```
tengqm, line 166: If we have a default value when declaring this parameter, we can reference `$CASE` directly.

.github/workflows/functional_tests_hetero_train.yml
```diff
@@ -0,0 +1,189 @@
...
+          else
+            echo "❌ Error: All 3 download attempts failed"
+            echo "Artifact name: ${{ inputs.source_artifact }}"
+            exit 1
+          fi
+
+      - name: Extract source code
+        run: |
+          mkdir -p $PROJECT_ROOT
+          tar -xzf /tmp/flagscale-source.tar.gz -C $PROJECT_ROOT
```
tengqm, line 103: same as above.

.github/workflows/functional_tests_hetero_train.yml
```diff
@@ -0,0 +1,189 @@
...
+          CASE='${{ matrix.test_config.case }}'
+          CONDA_ENV='${{ inputs.conda_env }}'
+          CONDA_PATH='${{ inputs.conda_path }}'
+
+          echo "Running functional tests for heterogeneous training"
+          echo "Platform: $PLATFORM"
+          echo "Device: $DEVICE"
+          echo "Task: $TASK"
+          echo "Model: $MODEL"
+          echo "Case: ${CASE:-all}"
```
tengqm, line 149: same as above

.github/workflows/unit_tests_common.yml
```diff
@@ -39,58 +54,114 @@ jobs:
...
+            fi
+          fi
+
+          # Run install script (handles conda activation internally if needed)
+          eval $INSTALL_CMD
+
+          # Copy test data (keep existing logic)
+          mkdir -p /opt/data
+          cp -r /home/gitlab-runner/data/Megatron-LM/* /opt/data/ || true
+          cp -r /home/gitlab-runner/tokenizers/Megatron-LM/* /opt/data/ || true
```
tengqm, line 128: how did we get this hard-coded paths?

tools/install/cuda/install_hetero_train.sh
```diff
@@ -0,0 +1,40 @@
...
+install_megatron_lm() {
+    local megatron_dir="$PROJECT_ROOT/Megatron-LM-FL"
+    local megatron_url="https://github.com/flagos-ai/Megatron-LM-FL.git"
+
+    log_info "Installing Megatron-LM-FL"
+
+    # Clone repository
+    retry_git_clone "$megatron_url" "$megatron_dir" "$RETRY_COUNT"
+
+    # Install from source
```
tengqm, line 27: This is hooking the stability of FlagScale to Megatron-LM-FL head, which is a moving target...
Ideally, we should pin to a latest stable tag and bump the tag when new releases of the Megatron-LM-FL is out.

tools/install/cuda/install_train.sh
```diff
@@ -0,0 +1,40 @@
...
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "$SCRIPT_DIR/../utils/utils.sh"
+source "$SCRIPT_DIR/../utils/retry_utils.sh"
+
+# Use inherited values or defaults for standalone execution
+PROJECT_ROOT="${PROJECT_ROOT:-$(get_project_root)}"
+RETRY_COUNT="${RETRY_COUNT:-3}"
+
+install_megatron_lm() {
```
tengqm, line 18: This is a duplication, and could be a potential risk for stability.

tools/install/utils/conda_utils.sh
```diff
@@ -0,0 +1,201 @@
...
+        echo "🐍 Found conda in PATH, activating environment: $env_name"
+        eval "$(conda shell.bash hook)"
+        conda activate "$env_name"
+        if [ $? -eq 0 ]; then
+            echo "✅ Successfully activated conda environment: $env_name"
+            return 0
+        fi
+    fi
+
+    # Method 2: Check common conda installation locations
```
tengqm, line 48: This smells an over design to me.
We may simply check if the `conda` command is available and fail fast if not.

tools/install/utils/conda_utils.sh
```diff
@@ -0,0 +1,201 @@
...
+    return 1
+}
+
+# Display conda and Python environment information
+# Usage: display_python_info
+display_python_info() {
+    echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
+    echo "  Python Environment Information"
+    echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
+
```
tengqm, line 94: We may want to remove things like this.

tools/install/utils/conda_utils.sh
```diff
@@ -0,0 +1,201 @@
...
+    if command -v python &> /dev/null; then
+        echo "Python location: $(which python)"
+        echo "Python version: $(python --version 2>&1)"
+    else
+        echo "⚠️  Python not found in PATH"
```
tengqm, line 99: Are we sure we want to proceed if python is not found?

tools/install/utils/conda_utils.sh
```diff
@@ -0,0 +1,201 @@
...
+        return 1
+    fi
+
+    # Get conda base directory
+    local conda_base=$(conda info --base)
+
+    if [ ! -f "$conda_base/bin/activate" ]; then
+        log_error "Conda activate script not found at $conda_base/bin/activate"
+        return 1
+    fi
```
tengqm, line 161: Why this is happening?

tools/install/utils/conda_utils.sh
```diff
@@ -0,0 +1,201 @@
...
+
+    log_step "Activating conda environment: $env_name"
+    source "$conda_base/bin/activate" "$env_name"
```
tengqm, line 164: Are there any output from this command?
If the activation fails, it simply fails silently?

tools/install/utils/load_platform_config.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+# Platform configuration loading script
+# Extracts platform configuration and groups tests by task
+# Usage: source load_platform_config.sh && load_platform_config <platform_name>
+
+set -euo pipefail
+
+load_platform_config() {
+    local PLATFORM=$1
+    local CONFIG_FILE=".github/configs/${PLATFORM}.yml"
```
tengqm, line 10: emm... shall we move this config file out of the `.github` directory?
I'm assuming that it is a file that can be used for other purposes.

tools/install/utils/load_platform_config.sh
```diff
@@ -0,0 +1,85 @@
...
+
+    echo "Loading platform configuration for: $PLATFORM"
+
+    if [ ! -f "$CONFIG_FILE" ]; then
+        echo "❌ Error: Platform configuration file not found: $CONFIG_FILE"
+        return 1
+    fi
+
+    # Extract CI/CD configuration from .github/configs using yq
```
tengqm, line 19: How can we ensure that `yq` is available?

tools/install/utils/load_platform_config.sh
```diff
@@ -0,0 +1,85 @@
...
+        return 1
+    fi
+    if [ -z "$RUNNER_LABELS" ] || [ "$RUNNER_LABELS" = "null" ]; then
+        echo "❌ Error: runner_labels not found in $CONFIG_FILE"
+        return 1
+    fi
+
+    # Query device types from test platform configuration
+    echo "Querying device types for platform: $PLATFORM"
+    DEVICE_TYPES=$(python tests/test_utils/runners/parse_config.py --platform "$PLATFORM" --type device_types 2>/dev/null || echo '[]')
```
tengqm, line 38: oh no. In addition to `yq`, we have a hand written parser for the YAML?

tools/install/utils/load_platform_config.sh
```diff
@@ -0,0 +1,85 @@
...
+    echo "Device types: $DEVICE_TYPES"
+
+    # Discover functional test matrix (device/task/model/case combinations)
+    echo "Discovering functional test matrix..."
+    FUNCTIONAL_TEST_MATRIX="[]"
+
+    # Parse functional tests from platform configuration for each device type
+    if [ -f "tests/test_utils/runners/parse_config.py" ]; then
```
tengqm, line 46: why are we testing the existence of the tool since we are already using it on line 38?

tools/install/install.sh
```diff
@@ -0,0 +1,379 @@
...
+    # This is the source of truth for which tasks are supported on the platform
+    if [ -f "$parse_config" ] && command -v python >/dev/null 2>&1; then
+        # Use parse_config.py to get functional tests from platform config
+        # Extract task names (top-level keys) from the JSON output
+        while IFS= read -r task; do
+            if [ -n "$task" ]; then
+                tasks+=("$task")
+            fi
+        done < <(python "$parse_config" --platform "$PLATFORM" --type functional 2>/dev/null | \
+                 python -c "import sys, json; data = json.load(sys.stdin); print('\\n'.join(data.keys()))" 2>/dev/null || true)
```
tengqm, line 44: This logic is dirty ... we combine python program file, python program text and shell script to extract a field from the given JSON.

tools/install/install.sh
```diff
@@ -0,0 +1,379 @@
...
+    fi
+
+    # Fallback method: Get tasks from install scripts that exist
+    # This ensures tasks with install scripts but no tests yet are still valid
+    if [ -d "$SCRIPT_DIR/$PLATFORM" ]; then
```
tengqm, line 49: Is this really a fallback?
Smells to me that this code block is always executed.

tools/install/install.sh
```diff
@@ -0,0 +1,379 @@
...
+            if [ -f "$script" ]; then
+                task=$(basename "$script" | sed 's/^install_//' | sed 's/\.sh$//')
+                if [ "$task" != "base" ]; then
+                    # Add task if not already in array
+                    if [[ ! " ${tasks[@]} " =~ " ${task} " ]]; then
+                        tasks+=("$task")
+                    fi
+                fi
+            fi
+        done
```
tengqm, line 60: With a clean `pyproject.toml` file, this whole script can be reduced to something like:

```shell
pip install ".[core]"
if [ "$task" = "inference" ]; then
   pip install ".[inference]"
fi
```

PR #1076, opened 2026-01-21
[wip] serve in foreground

Darryl233:

### DIFF
--- a/flagscale/runner/runner_serve.py
+++ b/flagscale/runner/runner_serve.py
@@ -1056,7 +1056,7 @@ class SSHServeRunner(RunnerBase):
         else:
             self.inference_engine = None
             self.port = None
-        self.background = self.config.experiment.get("runner", {}).get("background", False)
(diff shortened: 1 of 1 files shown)


flagscale/runner/runner_serve.py
```diff
@@ -1056,7 +1056,7 @@ def __init__(self, config: DictConfig):
         else:
             self.inference_engine = None
             self.port = None
-        self.background = self.config.experiment.get("runner", {}).get("background", False)
+        self.background = self.config.experiment.get("runner", {}).get("background", True)
```
tengqm, line 1059: So the new logic is that we are running this in the background by default?

Darryl233: Previous logic is running in the background by default. The new logic adds an option to control it but don't change the previous default logic. For aliyun image we can just provide a config file with background option turning off. Finally, this is not a formal PR, I will submit a formal PR later to merge into the main branch

PR #1081, opened 2026-01-22
Fix outdated installation instructions for Emu3.5

Teeeio:
### PR Category
Serve

### PR Types
Bug Fixes | Docs | Deprecations

### PR Description
This PR updates the installation method to use the vLLM-FL backend instead of the now-removed patch mechanism, aligning with the v1.0.0-alpha.0 refactor that migrated hardware-specific support into dedicated plugin repositories. Specifically:


tengqm (approved):

tengqm:
@Teeeio Looks like we have a conflict to be resolved.

PR #1103, opened 2026-02-03
Feat/mix precision integration

1235v:
## Overview

This PR introduces an automated data-free mixed-precision quantization pipeline
for large language models. The pipeline traverses the model layer by layer,
evaluates quantization sensitivity via cosine similarity, and automatically
assigns each layer an optimal precision strategy — balancing model size and
accuracy without requiring calibration data.

---

## Features

- **Data-free sensitivity analysis**: Measures per-layer weight distortion
  under four strategies using cosine similarity (no calibration dataset needed).
- **Four quantization strategies**:
  - `Std-8bit` — standard symmetric INT8
  - `QuIP-4bit` — incoherence-processed 4-bit
  - `W4A16` — 4-bit weight, 16-bit activation (symmetric)
  - `W4A16_ASYM` — 4-bit weight, 16-bit activation (asymmetric)
- **Two-stage decision logic**: First selects the best 4-bit strategy among
  the three 4-bit options; then falls back to 8-bit if the sensitivity score
  exceeds the configured threshold.
- **Auto group config generation**: Outputs mixed-precision group configs
  compatible with downstream inference engines.
- **End-to-end hook simulation**: Validates the final config via forward-hook
  emulation before writing the compressed checkpoint.

---

## Validated Models

| Model | Type |
|-------|------|
| Qwen3-0.6B | Dense |
| Qwen3-8B | Dense |
| Qwen3-32B | Dense |
| Qwen3-30B-A3B | MoE |
| Meta-Llama-3.1-8B | Dense |

---

## Usage

```bash
cd FlagScale
python run.py \
    --config-path examples/qwen35/conf/inference \
    --config-name compress_mix \
    action=run \
    compress.model.model_path=/path/to/your/model \
    compress.system.save_dir=/path/to/output \
    +experiment.runner.deploy.use_fs_serve=true \
    +compress.data.path=null

tail -f outputs/qwen35_mix/compress_logs/host_0_localhost.output 
```


tengqm:
@1235v Thanks for your contribution. Please sign the CLA so that we can review your PR.

physics31415926 (approved):

PR #1129, opened 2026-03-03
enable flagcx installation in flagscale setup.py

mikethegoblin:
### PR Category

Others
### PR Types

Improvements
### PR Description

This PR adds the capability of building and installing FlagCX inside FlagScale's `setup.py`. Here's a summary of the changes made in this PR:
  - Add FlagCX as a git submodule at third_party/FlagCX (pinned to v0.9.0)
  - Integrate FlagCX native library build and torch plugin installation into setup.py, controlled by pip extra arguments
  - Add several unit tests covering the new build integration
  
**Motivation**
FlagCX is a unified communication library that can be used with FlagScale for distributed training or inference. Previously, users had to clone and build FlagCX separately before using FlagScale. This change embeds FlagCX as a submodule and auto-builds it during pip install, similar to how PyTorch auto-builds NCCL.

**Usage**
The build is opt-in via pip extras:
e.g.
```bash
pip install ".[cuda, train, flagcx]"
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,33 +16,21 @@ from parse_requirements import parse_requirements
 def build_extras():
     """Build extras_require by scanning requirements/ directory.
 
-    Auto-discovers platforms (cuda, rocm, ...) and tasks (train, serve, ...).
-    Maps: requirements/<platform>/<task>.txt -> extra "<platform>-<task>"
-    Special: base.txt -> extra "<platform>", dev.txt -> extra "dev"
+    Produces flat extras: platform names (cuda, rocm, ...) and task names
+    (train, serve, ...) are separate extras with empty deps.  The actual
+    dependencies are installed via subprocess at install time because the
+    correct requirements file depends on the combination of platform + task,
+    which pip's static extras_require can't express.
 
-    Returns (extras, pip_options, pkg_options) where:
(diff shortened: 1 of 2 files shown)


tengqm:
I believe that there are some efforts on building deb/rpm packages for FlagCX ... what's the direction going forward?

mikethegoblin:
> I believe that there are some efforts on building deb/rpm packages for FlagCX ... what's the direction going forward?

I think these are not totally conflicting approaches. Users can still install FlagCX from deb/rpm packages; they can also choose to use the approach implemented in this PR for quick startup. Pytorch also allows users to use the NCCL installed on the system instead of the NCCL bundled with Pytorch.

setup.py
aoyulong, line 1: In the refactoring of the install process, we try hard not use the env var to control the install dependency. Please try to use another way instead of using the env var.In the refactoring of the installation process, we strive to avoid using environment variables to control the installation dependencies. Please consider using an alternative method instead of relying on environment variables.

mikethegoblin: I improved upon the initial version to use pip extras for triggering flagcx installation, e.g. `pip install [.flagcx-nvidia]`. This approach is consistent with the new installation method of FlagScale.

PR #1131, opened 2026-03-04
[CICD] refactor: migrate test runner from python run.py to flagscale CLI

zihugithub:
### PR Category
CICD

### PR Types
Improvements

### PR Description
- Replace old `python run.py` invocations with `flagscale` CLI commands
- Use bash arrays instead of eval for validator and get_test_configs
- Update serve stop to use `flagscale serve --stop`
- Simplify exp_dir extraction with grep -oP

### DIFF
--- a/.github/workflows/functional_tests_hetero_train.yml
+++ b/.github/workflows/functional_tests_hetero_train.yml
@@ -167,36 +167,6 @@ jobs:
             --no-system --no-dev --no-base --no-task \
             --src-deps megatron-lm \
             --retry-count 3
-        timeout-minutes: 30
-
-      - name: Install and verify FlagScale CLI
-        run: |
-          set -euo pipefail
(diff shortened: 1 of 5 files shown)


.github/workflows/functional_tests_hetero_train.yml
```diff
@@ -169,6 +169,43 @@ jobs:
             --retry-count 3
         timeout-minutes: 30
 
+      - name: Install and verify FlagScale CLI
```
Darryl233: Maybe it's better to merge this step into Install dependencies

zihugithub: Got it! I've merged this step into the 'Install dependencies' section as suggested.

Caozhou1995 (approved):
LGTM

PR #1197, opened 2026-05-08
[Fix] Temporarily fix missing mpu.get_engram_model_parallel_group

lzy-edu:
### PR Category

Train
### PR Types

Bug Fixes
### PR Description

- Fix missing `mpu.get_engram_model_parallel_group` function due to Megatron submodule version mismatch.
- Temporarily replace it with `mpu.get_engram_embedding_parallel_group` in `utils.py` to bypass the `calc_params_l2_norm` error.


aoyulong (approved):

LiJunscs:
image
The get_engram_model_parallel_gorup() is in the megatron-fl. There is no dismatch. And replace it with get_engram_embedding_parallel_group() will cause dismatch among different pipeline_stages.

lzy-edu:
Thanks, I have found the root cause of the problem. It happened because I used the official installation command but overlooked the megatron-lm version update.

PR #1221, opened 2026-06-11
fix pi05 pretrained loading memory peak

yilongma0110:
## Summary

This PR reduces the CPU memory peak during PI0.5 pretrained checkpoint loading.

Instead of constructing a fully initialized model on CPU before loading the checkpoint, the model is now built under `init_empty_weights()` and materialized with `load_state_dict(..., assign=True)`. After loading, the normal dtype setup, `requires_grad` setup, and `config.device` transfer are applied.

The code also checks that no meta parameters or buffers remain after checkpoint loading.

## Why fail fast on loading errors

The old path could safely return `model` after a checkpoint loading failure because `model = cls(config)` created real parameters.

With `init_empty_weights()`, the model starts as a meta-tensor skeleton. If checkpoint loading fails, returning it could return an incomplete model. Therefore this PR raises an error when pretrained loading cannot be completed safely.

## Test

In our Ascend 910B test, the PI0.5 model loaded successfully:

* Loaded `model.safetensors`
* Remapped 812 state dict keys
* All keys loaded successfully
* `pi0.5 model loaded successfully`

Observed CPU RSS peak decreased from about **18.19 GB** to **13.37 GB**.

A later `relative_actions_processor` registry error occurs after model loading and is unrelated to this pretrained-loading memory optimization.


legitnull (approved):

1016qqz (approved):
LGTM

PR #1227, opened 2026-06-17
[Model] Support to train glm5

heavyrain-lzy:
### PR Category

Train
### PR Types

Others
### PR Description

- Add GLM5/GLM5.1/GLM5.2 training support with new example configs under examples/glm5/, including both a small demo config and a 744B/40B Active model config.
- Add GLM5 checkpoint conversion support in tools/checkpoint/, but not verified enoughly.
- Update Megatron GPT builder to preserve experimental attention layer specs for MTP and print model structure/basic parameter-memory information after model construction.
- Fix checkpoint save flow so energy monitor pause/resume is only triggered when log_energy is enabled.


lxd-cumt (approved):
LGTM

zhaoyinglia (approved):

PR #1233, opened 2026-07-05
Fix ckpt convert for qwen35 moe model

wenqingqian:
fix ckpt convert for qwen35 moe model and support uneven pipeline parallel


zhaoyinglia (approved):

lxd-cumt (approved):
LGTM

PR #1243, opened 2026-07-21
Add low-overhead GPU progress heartbeat monitoring

lzy-edu:
### PR Category
Train
### PR Types
New Features

### PR Description
This PR adds a standalone, low-overhead GPU heartbeat monitoring capability for distributed training.

The main changes include:

- Record per-rank GPU progress heartbeats from existing Megatron fault-tolerance integration points.
- Run heartbeat recording on the CPU side without CUDA synchronization.
- Collect low-frequency, node-local GPU health information through `nvidia-smi`.
- Correlate each training rank with its assigned GPU.
- Integrate heartbeat lifecycle management into the training runner and SSH launcher.
- Write heartbeat status and findings as machine-readable JSONL logs.
- Add unit tests and two-worker integration scenarios.

The heartbeat distinguishes between:

- Rank process liveness.
- GPU training progress.
- Low-frequency GPU hardware health status.

#### Usage

Enable `heartbeat` under `experiment.runner` in the training configuration:

```yaml
experiment:
  runner:
    no_shared_fs: false
    heartbeat:
      enabled: true
      publish_interval_s: 5
      process_timeout_s: 30
      progress_timeout_s: 300
      checkpoint_timeout_s: 1800
      hardware_health:
        enabled: true
        interval_s: 60
```

After enabling the heartbeat, start the FlagScale training task with the existing launch command. No additional build step or separate monitoring command is required.

The runner automatically:

- Starts a lightweight heartbeat publisher for each training rank.
- Starts an optional GPU health collector on each node.
- Starts the heartbeat monitor on node 0.
- Stops the related monitoring processes when training finishes.

#### Results

Heartbeat monitoring results are written to:

```text
<log_dir>/heartbeat/<run_id>/
├── status.json
├── findings.jsonl
├── rank_<rank>*pid*<pid>.heartbeat.jsonl
└── gpu_health_node_<node_rank>.json
```

The expected behavior is:

- During normal training, each rank is reported as `alive`, and its GPU training progress is reported as `progressing`. No timeout finding is generated.
- After training completes normally, each rank is reported as `exited`, and its GPU progress is reported as `completed`.
- If a rank process or its heartbeat thread becomes unresponsive, a `rank_process_heartbeat_timeout` finding is recorded.
- If a rank process remains alive but its training progress stops advancing, a `gpu_progress_timeout` finding is recorded.
- When hardware health collection is enabled, the health status of the GPU assigned to the affected rank is included as supporting evidence.

For example, when a rank remains alive but its training progress stalls, `findings.jsonl` contains a record similar to:

```json
{
  "finding_type": "gpu_progress_timeout",
  "rank": 1,
  "process_liveness": "alive",
  "gpu_progress": "stalled",
  "reason": "rank_alive_but_gpu_progress_not_advancing"
}
```

These results make it possible to distinguish a rank process that has become unresponsive from a live rank whose GPU training progress has stalled, while also correlating the issue with optional GPU hardware health information.


flagscale/train/megatron/training/ft_integration.py
```diff
@@ -45,6 +45,8 @@
 
 import torch
 
+from flagscale.train import gpu_heartbeat
```
Darryl233, line 49: Better use relative imports like line50~line52

Darryl233 (dismissed):
LGTM

flagscale/runner/runner_train.py
```diff
@@ -303,13 +315,14 @@ def _generate_run_script_train(
             f.write(f'echo "Monitor service started in background for {host} (node {node_rank})"\n')
         f.write("\n")
 
+        command_body = heartbeat_config.training_command_body(node_rank)
```
zhaoyinglia, line 318: What happens when heartbeat_config.enabled is set to False?

flagscale/runner/launcher/launcher_ssh.py
```diff
@@ -148,6 +148,9 @@ def _get_runner_cmd_train(
         del runner_args["perf_max_log_files"]
     if "perf_model_type" in runner_args:
         del runner_args["perf_model_type"]
+    # Heartbeat is consumed by FlagScale and is not a torchrun option.
```
zhaoyinglia, line 151: add `NotImplementError` for CloudTrainRunner

flagscale/runner/heartbeat/monitor.py
```diff
@@ -0,0 +1,481 @@
...
+    ) -> list[dict[str, Any]]:
+        findings: list[dict[str, Any]] = []
+        for rank in range(self.expected_world_size):
+            if rank in self.ranks:
+                continue
+            age_s = now_s - self.monitor_started_s
+            key = (rank, -1, "initial_process_heartbeat")
+            if age_s <= self.initial_process_timeout_s or key in self._reported:
+                continue
+            self._reported.add(key)
```
zhaoyinglia, line 175: _reported uses (rank, pid, timeout_type) as the deduplication key and is never cleared when heartbeat or training progress resumes. As a result, if a rank stalls, recovers without restarting, and later stalls again, the second incident will not be written to findings.jsonl because the rank and PID remain unchanged.

flagscale/runner/heartbeat/health_reader.py
```diff
@@ -0,0 +1,210 @@
...
+        self.run_id = run_id
+        self.enabled = enabled
+        self.stale_after_s = stale_after_s
+        self._reported: set[tuple[str, str, tuple[str, ...]]] = set()
+
+    def poll(self, now_unix_ns: int) -> tuple[HardwareHealthIndex, list[dict[str, Any]]]:
+        if not self.enabled:
+            return HardwareHealthIndex(False, now_unix_ns, self.stale_after_s), []
+        snapshots: list[dict[str, Any]] = []
+        for path in sorted(self.heartbeat_dir.glob("gpu_health_node_*.json")):
```
zhaoyinglia, line 183: If there is a scene that error node dose not generate the gpu_health_node_*.json file?

flagscale/train/megatron/training/ft_integration.py
```diff
@@ -77,6 +78,10 @@ def get_rank_monitor_client() -> Optional[Any]:
 
 def setup() -> None:
     """Initialize fault tolerance before initialize_megatron"""
+    # GPU progress heartbeat is independent of the NVIDIA FT package. The runner
```
zhaoyinglia, line 84: add ##### FlagScale Begin ##### &&  ##### FlagScale END #####  comment

flagscale/runner/runner_train.py
lzy-edu: Thanks for pointing this out. When the heartbeat configuration is absent or heartbeat_config.enabled is set to False, prepare_heartbeat_launch_config() still returns a valid HeartbeatLaunchConfig(enabled=False) object. Therefore, heartbeat_config.training_command_body(node_rank) is still called safely.
In the disabled path, training_command_body() returns the original training command as $cmd; sync. The setup and stop command lists are empty, so the heartbeat publisher, monitor, and GPU health collector are not started.
We verified both scenarios with real two-GPU, 100-iteration training jobs:
      1. The heartbeat configuration was completely absent.
      2. The heartbeat configuration was present with enabled: false.
Both jobs completed 100/100 iterations, exited with code 0, and saved the iteration-100 checkpoint. No heartbeat processes or heartbeat/GPU-health artifacts were generated.

flagscale/runner/launcher/launcher_ssh.py
lzy-edu: I’ve added an explicit check for CloudTrainRunner. If heartbeat is enabled with the cloud runner, it now raises a clear NotImplementedError instead of entering an unverified execution path.
When heartbeat is disabled, it still follows the original no-op path and does not raise this error. I also added a unit test for this case.

flagscale/runner/heartbeat/monitor.py
lzy-edu: I’ve changed _reported so that it only deduplicates an active timeout, rather than permanently suppressing future timeouts for the same rank and PID.
When a valid heartbeat is received, I clear the corresponding process-heartbeat timeout key. When progress_seq advances, I clear the GPU-progress timeout key. I also clear old keys when a rank starts with a new PID.
The existing records in findings.jsonl are still preserved. With this change, the same process can now be reported correctly in this sequence: stall, recover, and stall again. I also added tests for both process-heartbeat and GPU-progress recovery scenarios.

flagscale/runner/heartbeat/health_reader.py
lzy-edu: I’ve added the expected node count to the heartbeat monitor, based on runner.nnodes.
After the startup grace period, the monitor checks whether every expected node has generated its gpu_health_node_*.json file. If a file is missing, I mark that node’s GPU health as unavailable and write a gpu_health_snapshot_missing finding containing the node rank and expected filename.
The same missing condition is reported only once while it remains active. If the file later appears, the missing state is cleared; if it disappears again, a new finding can be reported. I also added tests for startup grace, missing-file detection, recovery, and recurrence.

flagscale/train/megatron/training/ft_integration.py
lzy-edu: Done.

zhaoyinglia (approved):
LGTM

Darryl233 (approved):
LGTM

PR #1268, opened 2026-08-17
Fix process heartbeat false positives during checkpointing

lzy-edu:
### PR Category
Train

### PR Types
Bug Fixes

### PR Description
This PR fixes false rank_process_heartbeat_timeout findings during long-running checkpoints.
During checkpointing, GPU progress monitoring already uses checkpoint_timeout_s, but process-heartbeat liveness checks still use the regular process_timeout_s. A long checkpoint can temporarily block heartbeat publishing while the training process remains alive, causing a false timeout finding.
The main changes include:
1. Use checkpoint_timeout_s for subsequent process-heartbeat checks when the latest rank phase is checkpointing.
2. Continue using initial_process_timeout_s before the first heartbeat.
3. Continue using process_timeout_s during regular training phases.
4. Use the same phase-aware timeout logic for finding generation and status snapshots.
5. Record the detected phase and applied timeout in process-heartbeat findings.
6. Add tests for regular training, checkpoint grace, checkpoint timeout, and initial-heartbeat behavior.


flagscale/train/gpu_heartbeat.py
```diff
@@ -124,18 +134,56 @@ def mark_progress(self, phase: str, iteration: int | None = None) -> None:
             self._last_progress_unix_ns = now_unix_ns
             self._last_progress_mono_ns = now_mono_ns
 
+    def checkpoint_start(self, iteration: int | None = None) -> str:
```
Darryl233, line 137: This function will be called on every training step at flagscale/train/megatron/training/training.py:3460, which means every  training step will trigger a file write，and may introduce unacceptable overhead

lzy-edu: This has been fixed by passing the async-finalization and blocking context into the checkpoint hooks. Non-blocking per-step finalization polling no longer publishes any heartbeat checkpoint events. Actual save_checkpoint() operations are still recorded immediately, while blocking async finalization is recorded only when the async queue has pending work.

Darryl233 (approved):
lgtm

PR #1279, opened 2026-08-31
fix engram transformer integration

zhaoyinglia:
### PR Category

Train
### PR Types

Improvements
### PR Description

  Refactor and fix Engram integration for both the generic Engram model and DeepSeek V4.

  - Integrate Engram through TransformerLayer._forward_attention().
  - Reuse the standard Megatron TransformerBlock.
  - Share lazy asynchronous hash computation between Engram and DeepSeek V4.
  - Support platform-independent device streams.
  - Construct Engram modules only for configured Engram layers.
  - Preserve the residual connection around the Engram output.
  - Keep distributed checkpoint handling compatible with heterogeneous transformer layers.
  - Remove the duplicated custom TransformerBlock implementation.


lxd-cumt (dismissed):
LGTM

